PR #7, opened 2021-04-06
Add a "filter" to "/ExchangeSteps.ListExchangeSteps" gRPC method.

efoxepstein:
This also specifies that the Key is optional to allow searching across
multiple Exchanges at once.

These changes enable API clients to efficiently find work that needs to
be done (in other words, ExchangeSteps in the NOT_STARTED state).

world-federation-of-advertisers/cross-media-measurement#3

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/7)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto
@@ -29,6 +29,10 @@ service ExchangeSteps {
   rpc GetExchangeStep(GetExchangeStepRequest) returns (ExchangeStep);
 
   // Lists `ExchangeSteps`.
+  //
+  // This method is critical to the core "flow" of the system: clients will call
+  // "/ExchangeSteps.ListExchangeSteps" with a filter for the NOT_STARTED state
+  // and then create new `ExchangeStepAttempts` for those.
   rpc ListExchangeSteps(ListExchangeStepsRequest)
       returns (ListExchangeStepsResponse);
 }
@@ -41,7 +45,10 @@ message GetExchangeStepRequest {
 
 // Request message for `ListExchangeSteps` method.
 message ListExchangeStepsRequest {
-  // The Exchange that the ExchangeSteps belong to. Required.
+  // The Exchange that the ExchangeSteps belong to.
+  //
+  // Optional: if not provided, this will list ExchangeSteps from all Exchanges
+  // that the caller has access to.
   Exchange.Key parent = 1;
 
   // The maximum number of `ExchangeStep`s to return.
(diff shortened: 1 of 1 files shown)


oliver-amzn:
Why not make `parent` a part of filter as well? Looks to me it belongs in the same category.

efoxepstein:
Great question. We've been mostly following the AIPs (http://aip.dev) --
except that they're built for HTTP APIs so we do resource keys a bit
differently. One of the conventions in https://google.aip.dev/132 is that
the Request message has a "parent" field. (They use a string, we use Keys.)
There's also https://google.aip.dev/159, which discusses support for
wildcard parents. Does that make sense?

SanjayVas (approved):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/7)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @oliver-amzn)

src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto
```diff
@@ -60,6 +67,13 @@ message ListExchangeStepsRequest {
   // `ListExchangeStepsRequest` must match the call that provided the page
   // token.
   string page_token = 3;
+
+  // Filter for results. The API will return ExchangeSteps that match ALL of the
+  // given conditions. Repeated fields are treated as disjunctions.
+  message Filter {
+    repeated ExchangeStep.State states = 1;
+  }
+  Filter filter = 4;
```
kungfucraig, line 76: Are you sure you don't want repeated Filter to be the disjunction mechanism?

oliver-amzn:
Thx @efoxepstein  - makes sense!

oliver-amzn (approved):
🛰️

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/7)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto, line 76 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/7#-MXdIBVk8EImR0rANaW9-r1-76:-MXhaLdL9e-eUs_ZhJrf:bcfpyuq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/26a4ddfa476567da0acc6b5f73bd7651f21f3583/src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto#L76)):*

Are you sure you don't want repeated Filter to be the disjunction mechanism?
</blockquote></details>

It's a good idea, but this way is more consistent with existing Filters. See e.g. ListRequisitionsRequest or ListMeasurements. Luckily, I don't think we actually have any known use cases for such a disjunction.

kungfucraig (approved):

PR #9, opened 2021-04-12
Specify protocol config in MeasurementSpec instead of sketch config.

SanjayVas:
This additionally indicates which protocol should be used by the system. Unlike SketchConfig, protocol config is not a resource type.


wangyaopw (approved):

src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
```diff
@@ -161,15 +161,8 @@ message MeasurementSpec {
...
-    // Resource key of the sketch configuration. Required.
-    SketchConfig.Key sketch_config = 1;
-  }
-
-  // What data the requisition is for, i.e. what data is being requisitioned.
-  // Required.
-  oneof for {
-    // The data being requisitioned is an encrypted sketch.
-    EncryptedSketch encrypted_sketch = 2;
+  // Configuration for the computation protocol.
```
kungfucraig, line 164: Which protocol to use is basically an implementation detail, and should not be exposed to the MC. I also don't think that the MC should be able to set arbitrary sketch parameters (it will prevent optimizations on EDP side), but rather choose from a few well known configurations (1 to begin with).

The privacy parameters would be nice to expose.

We should discuss this. I think there are some complicated trade-offs.

kungfucraig (changes requested):

src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
wangyaopw: The requisition proto needs these details since the EDP needs these parameters to generate the data. The GlobalComputation Proto also needs these details since the worker needs them for computations.

Both these fields in the requisition proto and GlobalComputation proto should come from a same place, in this case, the measurement proto.

We don't need to expose all the fields in the measurement proto to the MC I think, maybe we are able to hide something in the UI?

SanjayVas: So the problem statement is that we need to have some parameters fixed at Requisition creation time since all participants (EDPs, Duchies) ultimately need to be operating on the same protocol configuration.

The next problem statement is that any meaningful parameters for a Measurement need to be signed by the MC. The question is which parameters these are, vs. which ones anyone else can freely change without the MC's consent. Note that we can have the Kingdom validate the request to ensure that parameters fall within some bounds, and have the MC frontend provide some default sets of params to choose from.

My understanding is that the `LiquidLegionsV2Config` that Yao defined is the minimal params that need to exist in the Requisition. If these aren't params that the MC needs to sign off on, then we can define an even smaller set of params (e.g. a single enum value) in the Measurement and have the Kingdom generate the other params. The key is that the MC must implicitly trust the Kingdom to generate the right params (the EDPs and Duchies will still have a chance to weigh in, but the MCs won't).

wangyaopw: FYI, the privacy parameters, i.e., noise parameters, will also be added to the LiquidLegionsV2Config, soon in a following PR

wangyaopw: I added these config to the Report proto.
https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/17

Some of the fields can come from the MC, the others can just be from a kingdom local config files.

As a minimum, the Measurement proto in the API should contain a field to indicate what this measurement is going to measure. 

For example, 
message ReachAndFrequency{
    a list of parameters suitable to be set by the MC.
}
oneof Foo{
  ReachAndFrequency reach_and_frequency = x;
}

The kingdom will use parameters set in Measurement.ReachAndFrequency and add others from a local config file.

SanjayVas: Moving this PR to draft to make changes re:this feedback.

SanjayVas:
Dropping this in favor of #13.

PR #13, opened 2021-04-21
Specify higher-level parameters in MeasurementSpec rather than SketchConfig resource key.

SanjayVas:
This allows the Kingdom to select an appropriate protocol and well-known protocol configuration for a given Measurement, abstracting the protocol-specific details from Measurement creation.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)


SanjayVas:
This replaces #9.

src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
```diff
@@ -0,0 +1,49 @@
...
+  // Encryption public key for the `Measurement` that this `MeasurementSpec` is
+  // associated with. Required.
+  bytes measurement_public_key = 1;
+
+  message ReachAndFrequency {
+    // Differential privacy parameters for the global reach DP noise registers.
+    //
+    // Each of these registers contains: a random register ID which is outside
+    // of the bounds of the normal ID space, the same constant key indicating
+    // that the register is destroyed, and an arbitrary count value.
```
wangyaopw, line 36: I am not sure if such details are appropriate in the api documentation.

@kungfucraig how should we describe these params to MC?

src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
```diff
@@ -0,0 +1,49 @@
...
+
+option java_package = "org.wfanet.measurement.api.v2alpha";
+option java_multiple_files = true;
+option java_outer_classname = "MeasurementSpecProto";
+
+// Specification for a `Measurement`. Immutable.
+message MeasurementSpec {
+  // Encryption public key for the `Measurement` that this `MeasurementSpec` is
+  // associated with. Required.
+  bytes measurement_public_key = 1;
```
wangyaopw, line 29: is this the key provided by the MC and used by the aggregator to hide the result from the kingdom?

src/main/proto/wfa/measurement/api/v2alpha/requisition.proto
```diff
@@ -51,13 +52,17 @@ message Requisition {
   // Immutable.
   SignedData measurement_spec = 4;
```
wangyaopw, line 53: what is the usage of measurement_spec in requisition?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: 0 of 8 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @wangyaopw)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 29 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15d-r1-29:-MYqwQUX5-i-Wddq1Ylp:bhj1965) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L29)):*

is this the key provided by the MC and used by the aggregator to hide the result from the kingdom?
</blockquote></details>

Yes, and also the public part of the key used by the MC to encrypt the requisition specs for each EDP, along with the EDP's public key for the hybrid KEM.
___
*[src/main/proto/wfa/measurement/api/v2alpha/requisition.proto, line 53 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15_-r1-53:-MYqxFQ91hnOZn6bvhqY:bp7w2t4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L53)):*

what is the usage of measurement_spec in requisition?
</blockquote></details>

For consent signaling. It's part of the "requisition fingerprint" that the EDP signs to create the `data_provider_participation_signature`.

src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
```diff
@@ -0,0 +1,79 @@
...
+    // Differential privacy parameters for data provider noise registers.
+    // Required.  Immutable.
+    //
+    // Each of these registers contains a well-known constant register ID, and
+    // arbitrary key and count values.
+    DifferentialPrivacyParams data_provider_noise = 2;
+
+    // Sketch configuration, generated from `sketch_params`. Output-only.
+    // Immutable.
+    SketchConfig sketch_config = 3;
```
wangyaopw, line 59: Sorry I didn't raise this up during previous discussion.

I already moved the Sketch/SketchConfig proto into the AnySketch repo.
https://github.com/world-federation-of-advertisers/any-sketch/blob/main/src/main/proto/wfa/any_sketch/sketch.proto

For sure, the API definition shouldn't use proto defined outside, but it feels it is painful to convert a v2alpha.SketchConfig to anysketch.SketchConfig since there are so many distributions inside or maybe even more in the future.

The only uncertain part during converting a LiquidLegionsSketchParams proto to SketchConfig proto is the constant used to mod the hashed key
https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/loadtest/config/liquid_legions_sketch_config.textproto#L17

In other words, to make sure all EDPs get the same SketchConfig for LLV2, we only need to provide an extra int64 SamplingIndicatorSize in the requisition.

maybe replace this `SketchConfig sketch_config = 3;` with `int64 sampling_indicator_size = 3;` and delete the sketch_config.proto

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @wangyaopw)

___
*[src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto, line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15e-r1-59:-MYug6K57VEvrlxHfFIq:b8yquyz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L59)):*

Sorry I didn't raise this up during previous discussion.

I already moved the Sketch/SketchConfig proto into the AnySketch repo.
https://github.com/world-federation-of-advertisers/any-sketch/blob/main/src/main/proto/wfa/any_sketch/sketch.proto

For sure, the API definition shouldn't use proto defined outside, but it feels it is painful to convert a v2alpha.SketchConfig to anysketch.SketchConfig since there are so many distributions inside or maybe even more in the future.

The only uncertain part during converting a LiquidLegionsSketchParams proto to SketchConfig proto is the constant used to mod the hashed key
https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/loadtest/config/liquid_legions_sketch_config.textproto#L17

In other words, to make sure all EDPs get the same SketchConfig for LLV2, we only need to provide an extra int64 SamplingIndicatorSize in the requisition.

maybe replace this `SketchConfig sketch_config = 3;` with `int64 sampling_indicator_size = 3;` and delete the sketch_config.proto 
</blockquote></details>

I'm not terribly concerned with having to convert between API and external versions of proto messages, as that's sort of par for the course when it comes to having a gRPC API. The concern is mainly about how likely it is that separate EDP client implementations produce incompatible sketches. This includes multiple versions of the same client library.

If you think that `LiquidLegionsSketchParams` with the additional field is sufficient for all EDPs to generate compatible sketches, then I'm happy to switch to that.

Note that the API makes no assumptions about whether the client is using any particular library/service, whether it be some turn-key EDP or even AnySketch. It can assume they have access to those implementations as reference though.

src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
wangyaopw: > > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)_ status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @wangyaopw)
> 
> _[src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto, line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15e-r1-59:-MYug6K57VEvrlxHfFIq:b8yquyz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L59)):_
> 
> Previously, wangyaopw (Yao Wang) wrote…
> I'm not terribly concerned with having to convert between API and external versions of proto messages, as that's sort of par for the course when it comes to having a gRPC API. The concern is mainly about how likely it is that separate EDP client implementations produce incompatible sketches. This includes multiple versions of the same client library.
> 
> If you think that `LiquidLegionsSketchParams` with the additional field is sufficient for all EDPs to generate compatible sketches, then I'm happy to switch to that.
> 
> Note that the API makes no assumptions about whether the client is using any particular library/service, whether it be some turn-key EDP or even AnySketch. It can assume they have access to those implementations as reference though.

Yep, a sketchConfig is no better than a LiquidLegionsSketchParams + the additional field, they contain the exact same information inside.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @wangyaopw)

___
*[src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto, line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15e-r1-59:-MYuzqSx9Xj_2r-M541p:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L59)):*

> > _[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)_ status: 0 of 8 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @wangyaopw)
> 
> _[src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto, line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15e-r1-59:-MYug6K57VEvrlxHfFIq:b8yquyz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L59)):_
> 
> Previously, wangyaopw (Yao Wang) wrote…
> I'm not terribly concerned with having to convert between API and external versions of proto messages, as that's sort of par for the course when it comes to having a gRPC API. The concern is mainly about how likely it is that separate EDP client implementations produce incompatible sketches. This includes multiple versions of the same client library.
> 
> If you think that `LiquidLegionsSketchParams` with the additional field is sufficient for all EDPs to generate compatible sketches, then I'm happy to switch to that.
> 
> Note that the API makes no assumptions about whether the client is using any particular library/service, whether it be some turn-key EDP or even AnySketch. It can assume they have access to those implementations as reference though.

Yep, a sketchConfig is no better than a LiquidLegionsSketchParams + the additional field, they contain the exact same information inside. 
</blockquote></details>

Done.

wangyaopw (approved):
> Reviewed 4 of 8 files at r1, 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

kungfucraig (changes requested):
> Reviewed 1 of 8 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: all files reviewed, 1 unresolved discussion (waiting on @wangyaopw)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15d-r1-36:-MYvoKvXDS3BgxVEe0mT:b-piyulp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L36)):*

I am not sure if such details are appropriate in the api documentation.

@kungfucraig how should we describe these params to MC?
</blockquote></details>

Agreed. Something like:

"These are the privacy parameters for the reach."

Is probably sufficient. The rest is implementation detail. For example, we could change the protocol entirely and this field would still specify the output noise, but the specifics of how we arrive at it could change significantly.

Similar concern for the frequency case.

I'd consider renaming the variable to "reach_privacy_params" and "frequency_privacy_params".

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: all files reviewed, 1 unresolved discussion (waiting on @wangyaopw)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 36 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13#-MYqwOMtpj4az_R5v15d-r1-36:-MYvwq0cBGweZDkubBWk:bf02p8w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9ea3d39fc6ae080e640cda08cccbd5ebf76b142f/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L36)):*

Agreed. Something like:

"These are the privacy parameters for the reach."

Is probably sufficient. The rest is implementation detail. For example, we could change the protocol entirely and this field would still specify the output noise, but the specifics of how we arrive at it could change significantly.

Similar concern for the frequency case.

I'd consider renaming the variable to "reach_privacy_params" and "frequency_privacy_params". 
</blockquote></details>

Done.

For posterity, the previous comments and field names were copied from cross-media-measurement/pull/17.

kungfucraig (approved):
> Reviewed 3 of 3 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/13)* status: all files reviewed, 1 unresolved discussion (waiting on @oliver-amzn, @wangyaopw, and @zachcwc)

oliver-amzn (approved):
👍

PR #23, opened 2021-04-30
Add elliptic curve ID to LiquidLegionsV2 protocol config.

SanjayVas:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/23)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
@@ -47,11 +47,15 @@ message ProtocolConfig {
     LiquidLegionsSketchParams sketch_params = 1;
 
     // Differential privacy parameters for data provider noise registers.
-    // Required.  Immutable.
+    // Required. Immutable.
     //
     // Each of these registers contains a well-known constant register ID, and
     // arbitrary key and count values.
     DifferentialPrivacyParams data_provider_noise = 2;
+
+    // ID of the OpenSSL built-in elliptic curve. For example, 415 for the
+    // prime256v1 curve. Required. Immutable.
+    int32 elliptic_curve_id = 3;
   }
   // Configuration for the specific protocol. Required. Immutable.
   oneof protocol {


oliver-amzn (approved):
☀️

zachcwc (approved):

src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
```diff
@@ -47,11 +47,15 @@ message ProtocolConfig {
...
     // Differential privacy parameters for data provider noise registers.
-    // Required.  Immutable.
+    // Required. Immutable.
     //
     // Each of these registers contains a well-known constant register ID, and
     // arbitrary key and count values.
     DifferentialPrivacyParams data_provider_noise = 2;
+
+    // ID of the OpenSSL built-in elliptic curve. For example, 415 for the
+    // prime256v1 curve. Required. Immutable.
```
wangyaopw, line 57: Maybe add a comment, "currently only key size(what do you call this?) = 256 bit is supported." (our library is hardcoded to use 32+1 bytes as the size of a EC Point, It is easy to refactor the code, just low priority)

wangyaopw (approved):

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/23)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @wangyaopw)

___
*[src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto, line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/23#-MZZAj_nxxXa9bVCb4zr-r1-57:-MZnFuTvD47YesPst9Vp:bdbfxhl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f257538c458b75ca49e343df18359839b3d1e791/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L57)):*

Maybe add a comment, "currently only key size(what do you call this?) = 256 bit is supported." (our library is hardcoded to use 32+1 bytes as the size of a EC Point, It is easy to refactor the code, just low priority)
</blockquote></details>

The configs are provided by the Kingdom and we're also unlikely to switch to another curve. This is here more for completeness and to make it clear that every such ElGamal in the computation needs to be using the same curve.

If we still want this comment, we can add it to the internal version since that's the one that would be used to add a new config (as opposed to a public `Create` method).

kungfucraig (approved):

PR #34, opened 2021-07-07
Update exchange shared input/output to support urls

jonmolle:
The size of the blobs exchanged in the api steps is likely to be too
large to reasonably support directly passing via the API. On top of
that, we want to avoid having the Kingdom own the storage for some of
these steps. To avoid issues, this swaps the 'shared' values to string
from blob.

* Only updates the shared input/output map values from blob to string
* Implementation is done in another repo

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
@@ -51,7 +51,12 @@ message ExchangeStep {
   //
   // The values of this map are signed urls pointing to cloud-hosted SignedData.
   // Each SignedData's payload is a consists of concatenated encrypted values of
-  // uniform length.
+  // uniform length. The urls themselves are signed and encrypted to prevent
+  // a malicious Kingdom operator from having access to the storage.
+  //
+  // In the case where the step's output is expected to be written to remote
+  // storage the urls wil point to a blob containing another signed url, in the
+  // same format described above, for the step's output to be written to.
   //
   // When ExchangeStepAttempts are completed successfully,
   // `/ExchangeStepAttempts.FinishExchangeStepAttempt` is provided
@@ -59,7 +64,7 @@ message ExchangeStep {
   // the shared_outputs are (logically) aggregated into one map, and then
   // `shared_inputs` is the subset of that map consisting of labels from
   // `step.shared_input_labels`.
-  map<string, string> shared_inputs = 5;
+  map<string, bytes> shared_inputs = 5;
 
   enum State {
     STATE_UNSPECIFIED = 0;
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
@@ -49,8 +49,9 @@ message ExchangeStep {
   //
   // The keys of this map are the values of the `step.shared_input_labels` map.
   //
-  // The values of this map are serialized SignedData. Each SignedData's payload
-  // is a byte string of concatenated encrypted values of uniform length.
+  // The values of this map are signed urls pointing to cloud-hosted SignedData.
+  // Each SignedData's payload is a consists of concatenated encrypted values of
+  // uniform length.
   //
   // When ExchangeStepAttempts are completed successfully,
   // `/ExchangeStepAttempts.FinishExchangeStepAttempt` is provided
@@ -58,7 +59,7 @@ message ExchangeStep {
   // the shared_outputs are (logically) aggregated into one map, and then
   // `shared_inputs` is the subset of that map consisting of labels from
   // `step.shared_input_labels`.
-  map<string, bytes> shared_inputs = 5;
+  map<string, string> shared_inputs = 5;
 
   enum State {
     STATE_UNSPECIFIED = 0;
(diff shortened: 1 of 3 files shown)


efoxepstein (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jonmolle)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto, line 62 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34#-Me5kBfh7xaSbssB05V0:-Me5kBfi757HzPbOC_7v:b9epu3o) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/03a92f6f54a02ea9ba247638021cd0cd7e5fb1ae/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto#L62)):*
> ```protobuf
>   // `shared_inputs` is the subset of that map consisting of labels from
>   // `step.shared_input_labels`.
>   map<string, string> shared_inputs = 5;
> ```

Do we want the URLs themselves to be signed then encrypted so the Kingdom can't access them?
___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto, line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34#-Me5kHF03p_Hj8KxErD4:-Me5kHF03p_Hj8KxErD5:bom04ze) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/03a92f6f54a02ea9ba247638021cd0cd7e5fb1ae/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto#L63)):*
> ```protobuf
>   // `step.shared_input_labels`.
>   map<string, string> shared_inputs = 5;
> ```

Should `ExchangeStep` have a  field for outputs?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jonmolle)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto, line 63 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34#-Me5kHF03p_Hj8KxErD4:-Me5mAt_-JhzgYUqBMqL:b-nknxo8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/03a92f6f54a02ea9ba247638021cd0cd7e5fb1ae/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto#L63)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Should `ExchangeStep` have a  field for outputs?
</blockquote></details>

nevermind!

jonmolle:
> Do we want the URLs themselves to be signed then encrypted so the Kingdom can't access them?

If the contents are always encrypted, is there any level of safety this would provide that we're concerned about? There isn't a particular reason they can't be, since the clients are going to handle the actual file transfers here. It would prevent some malicious actions by the kingdom, if that's a consideration.

efoxepstein (commented):
Correct, we want to prevent a malicious Kingdom from even knowing where the files are stored.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34#-:-MeWPXAU7qNVCFEkHZ-_:b-k3v5cv)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @jonmolle)

jonmolle:
Reverted back to bytes to indicate that any value passed through is expected to be encrypted. Only the comments are updated as (so far) this change is invisible to the API

efoxepstein (commented):
> Reviewed 3 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34)* status: all files reviewed, 2 unresolved discussions (waiting on @jonmolle)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto, line 54 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34#-MefA0BS49T8JaKcYUCN:-Mef9vVnEreDVdYTgDxw:b-39wnfp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/7720634f54d61985f32a7202e01a091c49f45a18/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto#L54)):*
> ```protobuf
>   // The values of this map are signed urls pointing to cloud-hosted SignedData.
>   // Each SignedData's payload is a consists of concatenated encrypted values of
>   // uniform length. The urls themselves are signed and encrypted to prevent
> ```

Can you elaborate on the format here? E.g. serialized SignedData protos where the `SignedData.data` fields, when interpreted as a UTF-8 string, are the URLs, for example.
___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 27 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/34#-MefAGUJ8W5BXWZu1uNW:-MefAGUJ8W5BXWZu1uNX:b3kre) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/7720634f54d61985f32a7202e01a091c49f45a18/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L27)):*
> ```protobuf
> //
> // See docs/panelmatch/example-exchange-workflow.textproto for an example
> // ExchangeWorkflow.S
> ```

?

PR #45, opened 2021-09-10
Remove `party` from ClaimReadyExchangeStepRequest

efoxepstein:
Instead, rely on an authentication mechanism to provide the principal DataProvider or ModelProvider.

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/45)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto
@@ -34,6 +34,9 @@ service ExchangeSteps {
   // If there are no ready ExchangeSteps, this will return an empty response.
   // Since this is expected, normal behavior, it does NOT return a NOT_FOUND
   // error.
+  //
+  // The caller must authenticate as either a DataProvider or a ModelProvider,
+  // and the returned ExchangeStep will belong to that party.
   rpc ClaimReadyExchangeStep(ClaimReadyExchangeStepRequest)
       returns (ClaimReadyExchangeStepResponse);
 }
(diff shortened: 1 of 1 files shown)


SanjayVas (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/45)* status: all files reviewed, 1 unresolved discussion (waiting on @efoxepstein)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto, line 39 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/45#-MjH2jvTAE3r_psjLACk:-MjH2jvTAE3r_psjLACl:b-2tdvav) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b5cc6b83c4b31caec2877b4b632a12c42415e39/src/main/proto/wfa/measurement/api/v2alpha/exchange_steps_service.proto#L39)):*
> ```
>   // The caller must authenticate as either a DataProvider or a ModelProvider,
>   // and the returned ExchangeStep will belong to that party.
> ```
`ExchangeStep` resources are those with multiple parents. The [canonical parent](https://google.aip.dev/124#multiple-many-to-one-associations) resource is the `Exchange`, and the secondary parent is the provider. Since `Claim` creates the `ExchangeStep`, the service needs to know both parents.

IIUC, the AIPs say to handle this is by specifying the parent resource name, but that the service may optionally allow a documented [resource ID alias](https://google.aip.dev/122#resource-id-aliases) such as `me` to be used determine the ID from the authenticated caller.

As an aside, `ExchangeStep` should probably have a field indicating the provider.

PR #46, opened 2021-09-13
Add Private Membership ExchangeWorkflow step types.

efoxepstein:
https://rally1.rallydev.com/#/?detail=/userstory/608878034535&fdp=true

---
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)

### DIFF
--- a/docs/panelmatch/example-exchange-workflow.textproto
+++ b/docs/panelmatch/example-exchange-workflow.textproto
@@ -8,6 +8,26 @@
 # * Each step is run on hardware controlled by the `party` given in each step. #
 ################################################################################
 
+# EDP generates a pepper used in generating AES keys.
+steps {
+  step_id: "generate-hkdf-pepper"
+  party: DATA_PROVIDER
+
+  input_step {}
+
+  shared_output_labels { key: "input" value: "hkdf-pepper" }
+}
+
+# EDP generates a pepper used in hashing encrypted identifiers.
+steps {
+  step_id: "generate-identifier-hash-pepper"
+  party: DATA_PROVIDER
+
+  input_step {}
+
+  shared_output_labels { key: "input" value: "identifier-hash-pepper" }
+}
+
 # MP provides private key as input.
 steps {
   step_id: "mp-input-commutative-deterministic-key"
@@ -99,7 +119,6 @@ steps {
   # appeared in the previous protocol.
   intersect_and_validate_step {
     max_size: 10000
-    min_size: 1000
     minimum_overlap: 0.99
   }
 
@@ -153,16 +172,54 @@ steps {
   private_output_labels { key: "output" value: "lookup-keys" }
 }
 
-# MP performs the lookup to get event-level data.
+# MP encrypts their queries.
 steps {
-  step_id: "lookup"
+  step_id: "prepare-event-lookup-queries"
   party: MODEL_PROVIDER
 
-  table_query_step {}
+  build_private_membership_queries_step {
+    serialized_parameters: "..."
+  }
 
-  private_input_labels { key: "keys" value: "lookup-keys" }
+  private_input_labels { key: "identifiers" value: "lookup-keys" }
 
-  shared_input_labels { key: "table" value: "encrypted-event-data" }
+  private_output_labels { key: "serialized-rlwe-private-key" value: "serialized-rlwe-private-key" }
+  private_output_labels { key: "raw-queries" value: "raw-queries" }
 
-  private_output_labels { key: "output" value: "encrypted-results" }
+  shared_output_labels { key: "encrypted-queries" value: "encrypted-queries" }
+  shared_output_labels { key: "serialized-rlwe-public-key" value: "serialized-rlwe-public-key" }
+}
+
+# EDP executes private membership queries
+steps {
+  step_id: "execute-private-membership-queries"
+  party: DATA_PROVIDER
+
+  execute_private_membership_queries_step {
+    serialized_parameters: "..."
+  }
+
+  private_input_labels { key: "event-data" value: "event-data" }
+
+  shared_input_labels { key: "encrypted-queries" value: "encrypted-queries" }
+  shared_input_labels { key: "serialized-rlwe-public-key" value: "serialized-rlwe-public-key" }
+
+  shared_output_labels { key: "encrypted-results" value: "encrypted-results" }
+}
+
+# MP decrypts the results
+steps {
+  step_id: "decrypt-private-membership-results"
+  party: MODEL_PROVIDER
+
+  decrypt_private_membership_query_results_step {
+    serialized_parameters: "..."
+  }
+
+  private_input_labels { key: "raw-queries" value: "raw-queries" }
+  private_input_labels { key: "serialized-rlwe-private-key" value: "serialized-rlwe-private-key" }
+
+  shared_input_labels { key: "encrypted-results" value: "encrypted-results" }
+
+  private_output_labels { key: "results" value: "results" }
 }
--- a/docs/panelmatch/example-exchange-workflow.textproto
+++ b/docs/panelmatch/example-exchange-workflow.textproto
@@ -8,7 +8,7 @@
 # * Each step is run on hardware controlled by the `party` given in each step. #
 ################################################################################
 
-# EDP generates a pepper used in generating AES keys.
+# EDP provides a pepper used in generating AES keys.
 steps {
   step_id: "generate-hkdf-pepper"
   party: DATA_PROVIDER
@@ -18,7 +18,7 @@ steps {
   shared_output_labels { key: "input" value: "hkdf-pepper" }
 }
 
-# EDP generates a pepper used in hashing encrypted identifiers.
+# EDP provides a pepper used in hashing encrypted identifiers.
 steps {
   step_id: "generate-identifier-hash-pepper"
   party: DATA_PROVIDER
@@ -181,10 +181,11 @@ steps {
     serialized_parameters: "..."
   }
 
+  private_input_labels { key: "identifiers" value: "raw-joinkeys" }
   private_input_labels { key: "identifiers" value: "lookup-keys" }
 
   private_output_labels { key: "serialized-rlwe-private-key" value: "serialized-rlwe-private-key" }
-  private_output_labels { key: "raw-queries" value: "raw-queries" }
+  private_output_labels { key: "query-decryption-keys" value: "query-decryption-keys" }
 
   shared_output_labels { key: "encrypted-queries" value: "encrypted-queries" }
   shared_output_labels { key: "serialized-rlwe-public-key" value: "serialized-rlwe-public-key" }
@@ -216,10 +217,11 @@ steps {
     serialized_parameters: "..."
   }
 
-  private_input_labels { key: "raw-queries" value: "raw-queries" }
+  private_input_labels { key: "query-decryption-keys" value: "query-decryption-keys" }
   private_input_labels { key: "serialized-rlwe-private-key" value: "serialized-rlwe-private-key" }
 
   shared_input_labels { key: "encrypted-results" value: "encrypted-results" }
+  shared_input_labels { key: "serialized-rlwe-public-key" value: "serialized-rlwe-public-key" }
 
   private_output_labels { key: "results" value: "results" }
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
+++ b/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
@@ -192,7 +192,6 @@ proto_library(
     srcs = ["exchange_step.proto"],
     strip_import_prefix = IMPORT_PREFIX,
     deps = [
-        ":crypto_proto",
         "@com_google_googleapis//google/api:resource_proto",
     ],
 )
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
@@ -17,7 +17,6 @@ syntax = "proto3";
 package wfa.measurement.api.v2alpha;
 
 import "google/api/resource.proto";
-import "wfa/measurement/api/v2alpha/crypto.proto";
 
 option java_package = "org.wfanet.measurement.api.v2alpha";
 option java_multiple_files = true;
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_step_attempts_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_step_attempts_service.proto
@@ -122,13 +122,4 @@ message FinishExchangeStepAttemptRequest {
 
   // Final debug log entries to append concurrently with the state transition.
   repeated ExchangeStepAttempt.DebugLog log_entries = 3;
-
-  // See documentation on `ExchangeStep.shared_inputs`.
-  //
-  // Contains an entry for each output accessible in the ExchangeStep's
-  // `step.shared_output_labels` map. The keys here are values from
-  // `shared_output_labels`.
-  //
-  // This field will be ignored if the `final_state` is `FAILED`.
-  map<string, bytes> shared_outputs = 4;
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto
@@ -49,12 +49,15 @@ message ExchangeWorkflow {
     // Inputs to the protocol are provided via InputStep step types -- this
     // ensures that all inputs to all steps appear as some output label.
     map<string, string> private_input_labels = 3;
+
     // Inputs to be fetched from shared storage
-    map<string, string> shared_input_labels = 11;
+    map<string, string> shared_input_labels = 4;
+
     // Outputs not shared
-    map<string, string> private_output_labels = 4;
+    map<string, string> private_output_labels = 5;
+
     // Outputs that are shared
-    map<string, string> shared_output_labels = 12;
+    map<string, string> shared_output_labels = 6;
 
     // Compares the inputs to determine if they overlap sufficiently.
     message IntersectAndValidateStep {
@@ -64,9 +67,6 @@ message ExchangeWorkflow {
       // The minimum fraction of items that must exist in both input data sets.
       float minimum_overlap = 4;
     }
-
-    // Selects a subset of key-value pairs.
-    message TableQueryStep {}
 
     // Applies deterministic, commutative encryption to the input plaintext
     // data.
@@ -86,15 +86,39 @@ message ExchangeWorkflow {
     // their inputs are ready.
     message InputStep {}
 
+    // Executes Private Membership queries.
+    message ExecutePrivateMembershipQueriesStep {
+      // This is a serialized instance of private_membership.batch.Parameters.
+      bytes serialized_parameters = 1;
+    }
+
+    // Builds Private Membership queries.
+    message BuildPrivateMembershipQueriesStep {
+      // This is a serialized instance of private_membership.batch.Parameters.
+      bytes serialized_parameters = 1;
+    }
+
+    // Decrypts results from Private Membership queries.
+    message DecryptPrivateMembershipQueryResultsStep {
+      // This is a serialized instance of private_membership.batch.Parameters.
+      bytes serialized_parameters = 1;
+    }
+
     oneof step {
-      IntersectAndValidateStep intersect_and_validate_step = 5;
-      TableQueryStep table_query_step = 6;
-      EncryptStep encrypt_step = 7;
-      ReEncryptStep reencrypt_step = 8;
-      DecryptStep decrypt_step = 9;
-      InputStep input_step = 10;
+      IntersectAndValidateStep intersect_and_validate_step = 7;
+      EncryptStep encrypt_step = 8;
+      ReEncryptStep reencrypt_step = 9;
+      DecryptStep decrypt_step = 10;
+      InputStep input_step = 11;
+      ExecutePrivateMembershipQueriesStep
+          execute_private_membership_queries_step = 12;
+      BuildPrivateMembershipQueriesStep build_private_membership_queries_step =
+          13;
+      DecryptPrivateMembershipQueryResultsStep
+          decrypt_private_membership_query_results_step = 14;
     }
   }
+
   // Sequence of steps of the workflow.
   repeated Step steps = 1;
 }


SanjayVas (changes requested):
> Reviewed 4 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @stevenwarejones)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjW9ZH21dCMMUaa6Cun:-MjW9ZH21dCMMUaa6Cuo:b-9pd5m6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L19)):*
> ```protobuf
> package wfa.measurement.api.v2alpha;
> 
> import "private_membership/rlwe/batch/cpp/shared.proto";
> ```

Adding some external proto to a public API is commonly the wrong thing to do. The safest approach is to have this API version have its own copy. We can ship conversion utils as necessary.

Otherwise we're binding this API version to a specific version of that external message, such that it would be incorrect for this API e.g. to handle new optional fields added to that message. If we're okay with that, we need to document exactly which version we're using for this API version.

https://google.aip.dev/213
https://google.aip.dev/215

stevenwarejones (changes requested):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @efoxepstein)

___
*[WORKSPACE, line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjWDMfw3thrtlJJqCb_:-MjWDMfw3thrtlJJqCba:bkv7wcx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/WORKSPACE#L40)):*
> ```Starlark
> )
> 
> http_archive(
> ```

do we want to move this to common? Not a fan of having two hashes to update consistently across two repos.
___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 98 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjWFAZz2LwPstbCaRpl:-MjWFAZz2LwPstbCaRpm:bqokxdy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L98)):*
> ```protobuf
>     // Builds Private Membership queries.
>     message BuildPrivateMembershipQueriesStep {
>       private_membership.batch.Parameters parameters = 1;
> ```

can we just serialize these and not depend on the external proto?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjW9ZH21dCMMUaa6Cun:-MjWHeqw2-wzjDF1djJG:b-lfabfn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L19)):*

Adding some external proto to a public API is commonly the wrong thing to do. The safest approach is to have this API version have its own copy. We can ship conversion utils as necessary.

Otherwise we're binding this API version to a specific version of that external message, such that it would be incorrect for this API e.g. to handle new optional fields added to that message. If we're okay with that, we need to document exactly which version we're using for this API version.

https://google.aip.dev/213
https://google.aip.dev/215
</blockquote></details>

Do we prefer duplicating this over just taking in a serialized version of it? If `private-membership` is updated, I see no reason why we should need to update `c-m-m-api`.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @efoxepstein)

___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjW9ZH21dCMMUaa6Cun:-MjWLetT8fS8LtiUfDxG:b-bkgls6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L19)):*
> If private-membership is updated, I see no reason why we should need to update c-m-m-api.

I believe this would be necessary to follow the AIP guidance on [backwards compatibility](https://google.aip.dev/180) and [version stability](https://google.aip.dev/181) if we were to include a message from there in our public API. private-membership isn't bound by the same API version constraints and could make a change that warrants a new API version to accomodate.

> Do we prefer duplicating this over just taking in a serialized version of it? 

If the API host (Kingdom) won't ever deserialize it, then serialized is fine. It basically means it's a a blind pass-through, and that readers and writers need to coordinate with each other rather than relying on API versioning. If the API host would deserialize it, then it would follow the same constraints as if the message were referenced directly in the API.

efoxepstein (commented):
> Dismissed @stevenwarejones from a discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[WORKSPACE, line 40 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjWDMfw3thrtlJJqCb_:-MjWMTvF2fMhGacLwoEF:b-ruqkw2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/WORKSPACE#L40)):*

do we want to move this to common? Not a fan of having two hashes to update consistently across two repos.
</blockquote></details>

Obsolete
___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 19 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjW9ZH21dCMMUaa6Cun:-MjWNtn29POhp0yxplT2:bpuieym) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L19)):*

> If private-membership is updated, I see no reason why we should need to update c-m-m-api.

I believe this would be necessary to follow the AIP guidance on [backwards compatibility](https://google.aip.dev/180) and [version stability](https://google.aip.dev/181) if we were to include a message from there in our public API. private-membership isn't bound by the same API version constraints and could make a change that warrants a new API version to accomodate.

> Do we prefer duplicating this over just taking in a serialized version of it? 

If the API host (Kingdom) won't ever deserialize it, then serialized is fine. It basically means it's a a blind pass-through, and that readers and writers need to coordinate with each other rather than relying on API versioning. If the API host would deserialize it, then it would follow the same constraints as if the message were referenced directly in the API.
</blockquote></details>

OK, changing to a blind passthrough. The Kingdom will never need to deserialize it.
___
*[src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto, line 98 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjWFAZz2LwPstbCaRpl:-MjWOPr87GsxI9Hymq3d:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/bcc683b91157af6ce9596ca9bdde720f937ac1f4/src/main/proto/wfa/measurement/api/v2alpha/exchange_workflow.proto#L98)):*

can we just serialize these and not depend on the external proto?
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 3 of 3 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 8 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-Mj_2O0sBbgmTS-pCNJb:-Mj_2O0t0EN-c5fwTW64:bj4sg7u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L188)):*
> ```
>   shared_input_labels { key: "encrypted-queries" value: "encrypted-queries" }
> 
>   shared_output_labels { key: "encrypted-results" value: "encrypted-results" }
> ```

what about sharing the hkdf_pepper and the identifier_pepper here too?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 8 unresolved discussions (waiting on @efoxepstein and @stevenwarejones)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-Mj_2O0sBbgmTS-pCNJb:-Mj_4GsX2mbdeWbt8WTl:b-1lzin) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L188)):*

what about sharing the hkdf_pepper and the identifier_pepper here too?
</blockquote></details>

actually, hkdf pepper here but identifier pepper should be shared earlier like in the DBE, right?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 8 unresolved discussions (waiting on @stevenwarejones)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 156 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZNISc8BeUCwguuiQ1:-Mj_3-M52y_jpNMD0HVd:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L156)):*

`prepare-event-lookup-query`?
</blockquote></details>

Done.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 160 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZNqv20dMzLrVT7jtn:-Mj_2rQiChlgodLxt9hK:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L160)):*

serialized_parameters
</blockquote></details>

Done.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 167 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZN_Tb8aENnqj36xLi:-Mj_69AA9lhbPXXVjw5O:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L167)):*

`serialized-rlwe-private-key`?
Also - we should also include the public-key too. As Ben explained it to me, either could be used for encrypting.
</blockquote></details>

Done.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 188 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-Mj_2O0sBbgmTS-pCNJb:-Mj_6EycDH5LRdJ27FO9:b-yq15sd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L188)):*

actually, hkdf pepper here but identifier pepper should be shared earlier like in the DBE, right?
</blockquote></details>

Added sharing both of these as the first two steps. The order won't actually matter -- since they're both sources in the DAG, they'll be immediately ready to be executed.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 197 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZNtzS-j2IyS_z_nJ3:-Mj_6NjSBQkRWHQxwAYw:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L197)):*

ditto
</blockquote></details>

Done.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-Mj_6XSv2t-C82Hb07xr:b-niwype) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*

does this include the single-blinded-joinkeys? If so, I think we need a better name that includes that
</blockquote></details>

Maybe it should be a map from queryId to the AES decryption key. WDYT?
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 203 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZNupJ2iID-snUbwlD:-Mj_6XCM-aOvHl1XypyQ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L203)):*

ditto
</blockquote></details>

Done.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @efoxepstein and @SanjayVas)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-Mja-El7BCuon2UDWI_n:b-i57qul) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Maybe it should be a map from queryId to the AES decryption key. WDYT?
</blockquote></details>

Yeah - they will need something like that. That's probably the simplest. Should it also include their panelistKey. Really, the output should be panelistKey to decrypted-events so panelistKeys need to be some kind of input.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 11 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-Mj_zdE46q5e_f_JEhBN:-Mj_zdE46q5e_f_JEhBO:bp4tu0a) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/753592afbf13cc0cace785f7f9950977f2bab067/docs/panelmatch/example-exchange-workflow.textproto#L11)):*
> ```
> ################################################################################
> 
> # EDP generates a pepper used in generating AES keys.
> ```

i prefer `provides` over `generates`. The other existing steps also use that terminology. Additionally, an EDP might, for example, have already been using a rotated pepper internally for a while before they need to provide it.

Or do you think the actual cadence of key generation needs to be part of the protocol as well?

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 4 of 6 files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-Mja9Lne1cErTSUomcxa:b-oq3912) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*

Yeah - they will need something like that. That's probably the simplest. Should it also include their panelistKey. Really, the output should be panelistKey to decrypted-events so panelistKeys need to be some kind of input.
</blockquote></details>

Done, PTAL
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 11 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-Mj_zdE46q5e_f_JEhBN:-Mja7c-99BWKc-MrNUyJ:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/753592afbf13cc0cace785f7f9950977f2bab067/docs/panelmatch/example-exchange-workflow.textproto#L11)):*

i prefer `provides` over `generates`. The other existing steps also use that terminology. Additionally, an EDP might, for example, have already been using a rotated pepper internally for a while before they need to provide it.

Or do you think the actual cadence of key generation needs to be part of the protocol as well?
</blockquote></details>

Done.

stevenwarejones (approved):
> Reviewed 1 of 3 files at r2, 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: 5 of 6 files reviewed, 1 unresolved discussion (waiting on @efoxepstein and @SanjayVas)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-MjaEqC13dJQ72X9gtjy:bhgx8ym) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

Done, PTAL
</blockquote></details>

do you that conveys the triplets of <QueryId, JoinKey, PanelistKey>?

SanjayVas (approved):
> Reviewed 1 of 1 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 1 unresolved discussion (waiting on @efoxepstein)

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-MjagnJg-hClQ-u14zgY:bpqcz2z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*

do you that conveys the triplets of <QueryId, JoinKey, PanelistKey>?
</blockquote></details>

I feel like type information isn't really embedded in the other labels either. I'm not sure what the best name would be. What do you think?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 3 unresolved discussions (waiting on @efoxepstein)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-MjdVk2Z1a1ZcNMEGMM7:b-ey5pdo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*
<details><summary><i>Previously, efoxepstein (Eli Fox-Epstein) wrote…</i></summary><blockquote>

I feel like type information isn't really embedded in the other labels either. I'm not sure what the best name would be. What do you think?
</blockquote></details>

i don't know. I'm kind of okay with the current name but maybe something more generic like "query-results-decryption-inputs" or something
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 191 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjdVcw16CBiShh7ccR0:-MjdVcw16CBiShh7ccR1:bqp4myx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/0e5f2cedbab7db611b03ba3b02f7e7cd203e437c/docs/panelmatch/example-exchange-workflow.textproto#L191)):*
> ```
> 
>   shared_output_labels { key: "encrypted-queries" value: "encrypted-queries" }
>   shared_output_labels { key: "serialized-rlwe-public-key" value: "serialized-rlwe-public-key" }
> ```

see below. Maybe write this public key to both private and shared.
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 224 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjdULMJ9FmWiKTJGBSU:-MjdULMJ9FmWiKTJGBSV:b-ketwrd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/0e5f2cedbab7db611b03ba3b02f7e7cd203e437c/docs/panelmatch/example-exchange-workflow.textproto#L224)):*
> ```
> 
>   shared_input_labels { key: "encrypted-results" value: "encrypted-results" }
>   shared_input_labels { key: "serialized-rlwe-public-key" value: "serialized-rlwe-public-key" }
> ```

This should really just come from their private input right? They generated it. I normally think of shared as what you get from the other party and you verify it was signed by them not by you.

efoxepstein (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: all files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[docs/panelmatch/example-exchange-workflow.textproto, line 202 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjZO4rW1An2BUGJ0fCf:-MjeDNrG1y-XlK-dnm5y:b-gwj81s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/401232781e5a18d7535ee8a89fcab2b61d896e3b/docs/panelmatch/example-exchange-workflow.textproto#L202)):*

i don't know. I'm kind of okay with the current name but maybe something more generic like "query-results-decryption-inputs" or something
</blockquote></details>

Let's leave it as is for now -- this is just an example workflow and we can always change it later. I'm not super happy with "query-results-decryption-inputs"
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 191 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjdVcw16CBiShh7ccR0:-MjeCxBME9x7ZqXmDaLQ:b-dt6t2z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/0e5f2cedbab7db611b03ba3b02f7e7cd203e437c/docs/panelmatch/example-exchange-workflow.textproto#L191)):*

see below. Maybe write this public key to both private and shared.
</blockquote></details>

Can we revisit this after we discuss my idea of an explicit "sharing" step type?
___
*[docs/panelmatch/example-exchange-workflow.textproto, line 224 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46#-MjdULMJ9FmWiKTJGBSU:-MjeD-vtCch_ryYcE1zd:b-dt6t2z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/0e5f2cedbab7db611b03ba3b02f7e7cd203e437c/docs/panelmatch/example-exchange-workflow.textproto#L224)):*

This should really just come from their private input right? They generated it. I normally think of shared as what you get from the other party and you verify it was signed by them not by you.
</blockquote></details>

Can we revisit this after we discuss my idea of an explicit "sharing" step type?

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/46)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @efoxepstein)

PR #81, opened 2021-12-15
Rename DeleteApiKey to RevokeApiKey.

tristanvuong2021:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/81)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/api_keys_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/api_keys_service.proto
@@ -31,11 +31,11 @@ service ApiKeys {
   // `MeasurementConsumer` the `ApiKey` is being created for.
   rpc CreateApiKey(CreateApiKeyRequest) returns (ApiKey) {}
 
-  // Deletes an `ApiKey`.
+  // Revokes an `ApiKey`.
   //
   // Results in PERMISSION_DENIED if the authenticated caller does not own the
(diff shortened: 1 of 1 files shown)


SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/81)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @tristanvuong2021)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/81#-MqzjdM41LI5IYSgggv-:-MqzjdM41LI5IYSgggv0:b-1dqxfs) (no related file):*
Broader question: why are we setting a revocation state instead of just deleting?

---

tristanvuong2021:
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/81#-MqzjdM41LI5IYSgggv-:-MqzyVX4ACDX9zpFhRo7:b-dfetvq) (no related file):*

Broader question: why are we setting a revocation state instead of just deleting?
</blockquote></details>

After seeing this and thinking about it, I realized my original reasoning is flawed. I was thinking what if it is deleted and then another one is made with the same auth key. But now I realized that doesn't matter. It would either work for a different measurement consumer, in which case attempts to use it with the old measurement consumer would fail, or it would be for the same measurement consumer, in which case nothing changes.

---

PR #86, opened 2022-02-01
Updates measurement_spec to include vid sampling interval.

matthewclegg:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -45,15 +45,15 @@ message MeasurementSpec {
     float width = 2;
   }
 
-  // Range of VIDs that will be included in this measurement
-  VidSamplingInterval vid_sampling_interval = 4;
-
   message ReachAndFrequency {
     // Differential privacy parameters for reach.
     DifferentialPrivacyParams reach_privacy_params = 1;
 
     // Differential privacy parameters for frequency.
     DifferentialPrivacyParams frequency_privacy_params = 2;
+
+    // Range of VIDs that will be included in this measurement
+    VidSamplingInterval vid_sampling_interval = 4;
   }
 
   // Fields specific to the type of measurement.
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -35,6 +35,19 @@ message MeasurementSpec {
   // `Measurement`. Required.
   repeated bytes nonce_hashes = 2;
 
+  // Specifies a range of VIDs to be sampled.
+  // If start + width <= 1, then the sampled vids are those
+  // satisfying 0.0 <= start < vid < start + width.
+  // If start + width > 1, then the sampled vids are those
+  // that satisfy (start <= vid) or (vid < start + width - 1)
+  message VidSamplingInterval {
+    float start = 1;
+    float width = 2;
+  }
+
+  // Range of VIDs that will be included in this measurement
+  VidSamplingInterval vid_sampling_interval = 4;
+
   message ReachAndFrequency {
     // Differential privacy parameters for reach.
     DifferentialPrivacyParams reach_privacy_params = 1;


wangyaopw (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @matthewclegg and @SanjayVas)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-MuqU7fs1aEq7NtiHPdD:-MuqU7fs1aEq7NtiHPdE:bmahtuu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/1570045c46142dd722a256d28067536d6dc29fe7/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L49)):*
> ```protobuf
> 
>   // Range of VIDs that will be included in this measurement
>   VidSamplingInterval vid_sampling_interval = 4;
> ```

2 issues

1, I assume this field only applys to a subset of measurements. For example, if we are measuring the total impression counts, there is no need to do sampling at all. So likely this should be defined inside protocols that require sampling.

2. measurementSpec only contains fields that can be configured by the measurement consumer. This `samplingInterval` will be determined by the kingdom, so it shouldn't be included here. The public Api LiquidLegionsV2 protocolConfig seems to be the place to define this, since this is used by both the EDP and the duchy.
https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L48



_Code quote:_
```protobuf
VidSamplingInterval
```

matthewclegg:
Hi Yao,

Thank you very much for your comments, and so sorry that you are leaving Google.  In response to your comments:

(1) There is no harm in having the sampling interval outside of the ReachAndFrequency message.  Later we will be adding a ReachOnly message, which will require its own sampling interval.  It seems easiest to have the VID sampling interval code in one place rather than to duplicate it in multiple places.  With regard to the planned ImpressionCount request, the code for handling impression count requests is free to ignore the VID sampling interval.  However, we might actually want to use a sampling interval for this as well, as it would potentially allow for a greater number of queries that can be handled with a given privacy budget.  So, this issue comes down to a choice between two design approaches:  (a) duplicate the code for the VID sampling interval in every message type that needs it, or (b) only include the code once, with the understanding that the code for handling some message types may choose to disregard the sampling interval.  My vote is for (b), but I am happy to go with whatever the team prefers.

(2) The choice of the sampling interval is bound together with the choice of epsilon and delta.  The accuracy of a response and the amount of DP budget that is used depends jointly on all three quantities.  If the MC is able to choose epsilon and delta, they should also be able to choose the sampling interval.  Yes, I earlier proposed that epsilon and delta should be moved out of the MeasurementSpec and determined internally by the Kingdom, and I still believe that would be a better overall design.  However, for now I would like to defer that decision and to work with the current design.

Let me know what you would like me to do, and I will be happy to oblige.  Thanks!

wangyaopw (commented):
The current plan looks good to me. 

Btw, I guess you are replying in github and that is why your comment is not attached to my original comment. Try click the purple button "reviewable", we do all code reviews inside reviewable, you can easily reply to specific comment there.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-:-Mur1NJd559IpCcQK46s:bmwq83w)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @matthewclegg and @SanjayVas)

wangyaopw (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: all files reviewed, 1 unresolved discussion (waiting on @matthewclegg)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-Mus0MDz68HX8ZtD8oRq:-Mus0MDz68HX8ZtD8oRr:bvfm0qo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/1570045c46142dd722a256d28067536d6dc29fe7/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L49)):*
>  (a) duplicate the code for the VID sampling interval in every message type that needs it, or (b) only include the code once, with the understanding that the code for handling some message types may choose to disregard the sampling interval. My vote is for (b), but I am happy to go with whatever the team prefers.

The main reason to go with (a) is so we can mark the field as required within the appropriate measurement types. (b) is okay, but the field description should include the conditions under which the field is required to be set.

Note that we eventually want to switch to [field behavior annotations](https://google.aip.dev/203) for required fields where possible.

wangyaopw (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-Mus0MDz68HX8ZtD8oRq:-MusbhSME1uwFrc4xNsF:b6th58f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/1570045c46142dd722a256d28067536d6dc29fe7/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L49)):*

>  (a) duplicate the code for the VID sampling interval in every message type that needs it, or (b) only include the code once, with the understanding that the code for handling some message types may choose to disregard the sampling interval. My vote is for (b), but I am happy to go with whatever the team prefers.

The main reason to go with (a) is so we can mark the field as required within the appropriate measurement types. (b) is okay, but the field description should include the conditions under which the field is required to be set.

Note that we eventually want to switch to [field behavior annotations](https://google.aip.dev/203) for required fields where possible.
</blockquote></details>

I vote for (a)

matthewclegg (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-Mus0MDz68HX8ZtD8oRq:-MuvJU0JAWfZ8OfGFob5:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/1570045c46142dd722a256d28067536d6dc29fe7/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L49)):*

I vote for (a)
</blockquote></details>

Done.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: all files reviewed, 1 unresolved discussion (waiting on @matthewclegg)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-MvLwm8xCYR1krPzs07d:-MvLwm8xCYR1krPzs07e:bbrgvls) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/10f43103289446b6d4e63a1fa100ad5e4a740fdb/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L43)):*
> ```protobuf
>   // If start + width > 1, then the sampled vids are those
>   // that satisfy (start <= vid) or (vid < start + width - 1)
>   message VidSamplingInterval {
> ```

I have mixed feelings about making these floats.

matthewclegg (commented):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-:-MvS1r3m8il5fRZIdMu5:bnfp4nl)* status: all files reviewed, 1 unresolved discussion (waiting on @matthewclegg)

matthewclegg (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-MvLwm8xCYR1krPzs07d:-MvS2I5v9afJZd9wv1WN:b-tjlm7w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/15dfb0f3bd74c5b9a57356159bebcf32840fb841/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L43)):*

I have mixed feelings about making these floats.
</blockquote></details>

Per discussion in the RF & Planning workstream, we will stick with floats for now.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @matthewclegg)

kungfucraig:
___
*[src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto, line 43 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/86#-MvLwm8xCYR1krPzs07d:-MvUEYswDT8LKaLzH__C:b-3pivzt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/15dfb0f3bd74c5b9a57356159bebcf32840fb841/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L43)):*
<details><summary><i>Previously, matthewclegg (Matthew Clegg) wrote…</i></summary><blockquote>

Per discussion in the RF & Planning workstream, we will stick with floats for now.
</blockquote></details>

The trade off we talked about was this. If we use the interval the API is pretty resilient to changes in the number of buckets, adding query types that require different numbers of buckets, etc.

On the other hand, this requires that the client manage the use of the privacy budget to ensure that there are not inaccessible gaps in the budget. For example, if we have a query that needs 3/10 of the unit interval and we repeatedly allocate [0, 0.3) and [0.4, 0.7) then the interval [0.3, 0.4) becomes unusable. If this actually becomes an issues we could remedy it by allowing a list of non-overlapping intervals to specify the budget instead of just the single interval.

PR #113, opened 2022-09-29
Add boolean flag in protocol_config for direct reach and frequency measurement

iverson52000:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113)

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
@@ -70,6 +70,9 @@ message ProtocolConfig {
     // May only be set when `measurement_type` is `REACH_AND_FREQUENCY`.
     LiquidLegionsV2 liquid_legions_v2 = 3;
   }
+
+  // If only single DataProvider, set to true
+  bool is_direct_RF_measurement = 4;
 }
 
 // Parameters for a Liquid Legions sketch.
--- a/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto
@@ -56,7 +56,8 @@ message Requisition {
 
   // Denormalized `protocol_config` field from `measurement`. Immutable.
   //
-  // If this is set, the requisition can be fulfilled by calling
+  // If this is set and is_direct_RF_measurement flag is false, the requisition
+  // can be fulfilled by calling
   // RequisitionFulfillment/FulfillRequisition(`measurement_type` is
   // `REACH_AND_FREQUENCY` and it has multiple DataProviders)
   //
@@ -64,7 +65,8 @@ message Requisition {
   // calling Requisitions/FulfillDirectRequisition
   // (volumetric measurement - `measurement_type` is `IMPRESSION`/`DURATION`
   // or direct reach and frequency measurement - measurement_type` is
-  // `REACH_AND_FREQUENCY` and it only has single DataProvider)
+  // `REACH_AND_FREQUENCY` and is_direct_RF_measurement flag is set to true that
+  // it only has single DataProvider)
   ProtocolConfig protocol_config = 5;
 
   // Denormalized `data_provider_certificate` field from the corresponding


SanjayVas (changes requested):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000, @matthewclegg, @rmirisola, @robinsons, and @xnegron)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto` line 75 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113#-NDZHf_8BPy4vrzLG7YQ:-NDZHf_90I0k5Hw2OxNw:b-5vq8wj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f2e591bd25e69cbcd14fe06d9837da07a6662913/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L75)):*
protobuf field names are in `lower_snake_case`. The direct translation of this field name would be `is_direct_r_f_measurement` but this would result in incorrectly-cased[^1] generated Java/Kotlin code. This is why it's often better to avoid acronyms/initialisms unless they're well-established.

We also generally avoid the `is` prefix on fields[^2]: 

> Fields of type boolean are often named like boolean accessor methods with the initial `is` omitted

Having the `is` prefix on the proto field would result in the Java generated code having a `getIsFoo()` method with the Kotlin property being `isFoo`, where we'd really want the Kotlin property to just be `foo`.

[^1]: https://developer.android.com/kotlin/style-guide#camel_case. Note that this preempts the item in https://kotlinlang.org/docs/coding-conventions.html#choose-good-names re: two-letter acronyms.
[^2]: Effective Java, 3rd Edition, Item 68



_Code quote:_
```protobuf
is_direct_RF_measurement
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113)* status: all files reviewed, 2 unresolved discussions (waiting on @iverson52000, @matthewclegg, @rmirisola, @robinsons, and @xnegron)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto` line 74 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113#-NDZUg7iCJwJI0sIc_d_:-NDZUg7iCJwJI0sIc_da:b-qd92fw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f2e591bd25e69cbcd14fe06d9837da07a6662913/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto#L74)):*
> ```protobuf
>   }
> 
>   // If only single DataProvider, set to true
> ```

This doesn't seem quite right. I thought that the boolean field is meant to abstract this logic away.

If the EDP is meant to be able to choose, then perhaps this should be a field that indicates that the requisition *could* be fulfilled directly. e.g. `directly_fulfillable`. Though then the EDP is still stuck having to check two places to determine whether to fulfill a requisition directly: this and the measurement type in `MeasurementSpec`.

We should have a simple way for the EDP to determine what they must/may do. Perhaps a repeated message containing the protocol one-ofs, where each item is one of the protocols that the EDP may use. One of these protocols can be "direct", which would be included either when there's only one EDP in an R/F measurement and the only option when it's a volumetric measurement.



_Code quote:_
```protobuf
// If only single DataProvider, set to true
```

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113)* status: all files reviewed, 3 unresolved discussions (waiting on @iverson52000, @matthewclegg, @rmirisola, @robinsons, and @xnegron)

*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/113#-NDZY8dVEfgHWTp09M0d:-NDZY8dW55-Zt5-tjbc6:b-1d8nok) (no related file):*
Make sure to update the documentation for the `protocol_config` field in `Requisition` accordingly.

___

PR #123, opened 2023-03-07
add ErrorDetails message to measurement proto message

Marco-Premier:
Adding ErrorDetails proto message to public API. 
In case of a measurement failure, the error is not inside Measurement.Details anymore.

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
+++ b/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
@@ -163,6 +163,7 @@ proto_library(
         ":protocol_config_proto",
         "@com_google_googleapis//google/api:resource_proto",
         "@com_google_protobuf//:duration_proto",
+        "@com_google_protobuf//:timestamp_proto",
     ],
 )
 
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/123)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/123)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 171 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/123#-NPxXqSb9nytJQtWgFvI:-NPxXqSb9nytJQtWgFvJ:b2issx3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/92202f5c992a18dc75212efc98dd48886dc3090f/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L171)):*
> ```protobuf
>       COMPUTATION_PARTICIPANT_FAILED = 3;
>     }
>     Reason reason = 1;
> ```

This reason is still quite important, meaning we wouldn't want to change this API.

I think I misunderstood that you were speaking about the public API in the comment on #875. I had assumed we could determine this from the MeasurementLogEntries table, but perhaps we're not always persisting the cause of the transition aside from `COMPUTATION_PARTICIPANT_FAILED`

Marco-Premier:
Closing this PR as public shouldn't change. Refer to #875

Issue #140, opened 2023-05-30
ClaimReadyExchangeStepRequest does not specify which resource or collection it is operating on.


SanjayVas:
This is similar to #138.

The `ClaimReadyExchangeStep` method appears to operate on a resource collection (the same way that a standard Create method does). This means it needs a `parent` field referencing the parent resource. This request message only has a oneof `party` field.

Alternatively the method can be changed to operate on a single resource by `name`, using a separate method to find the resource to operate on (e.g. List). An `etag` field can be added as per [AIP-154](https://google.aip.dev/154) if needed to provide freshness validation.

The latter option is more flexible at the cost of needing to make two RPCs.

See [AIP-136](https://google.aip.dev/136)

- [x] #142 
- [x] https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1284

SanjayVas:
There's a complication here in that `Exchange`s and `ExchangeStep`s appear to be created lazily. An `Exchange` resource for a given date will never exist before that date. Moreover, this lazy initialization does not occur in the current implementation until `ClaimReadyExchangeStep` is called. Even then, an `Exchange` might not be created if there's still a claimable `ExchangeStep` on any `RecurringExchange` for the caller.

SanjayVas:
There's a further complication: while a `RecurringExchange` has two owners, an `ExchangeStep` descendant of that `RecurringExchange` only has one of these as its owner. A ready `ExchangeStep` is only claimable by this owner. This owner is not exposed in the public API.

This means that even if one were to use a resource name that indicated an appropriate parent of the `RecurringExchange`, the `ExchangeStep` claimed by `ClaimReadyExchangeStep` would still vary by the authenticated principal.

PR #141, opened 2023-05-30
Replace ClaimReadyExchangeStep with method that operates on a single resource.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/141)

SanjayVas:
For #140

PR #146, opened 2023-06-09
Run api-linter from lint action instead of custom workflow.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/146)

Issue #144, opened 2023-06-05
Some List method responses vary by authenticated user


SanjayVas:
From [AIP-132](https://google.aip.dev/132):

> **Note:** List methods **should** return the same results for any user that has permission to make a successful List request on the collection. Search methods are more relaxed on this.

Furthermore the `filter` field should never be required.

The current methods which do not comply with this are:
* `ListEventGroups`
  * Filtering on `measurement_consumer` is effectively required when listing across `DataProvider`s.
* `ListExchangeSteps`
  * Filtering on the `data_provider` or `model_provider` of the `RecurringExchange` is effectively required.
  
The general solution in these cases is to have the resource collection be a child of a resource that maps to the authenticated user. In the cases above, having additional non-canonical patterns for the `eventGroups` and `recurringExchanges` resources that are supported by the List methods, mentioning the the `child_type` in the resource reference for the `parent` field in the List request.

In the case of `EventGroup`s, this means that `EventGroup` IDs must also be unique to each `MeasurementConsumer`.

PR #148, opened 2023-06-12
Add a secondary EventGroup pattern with MeasurementConsumer as the parent.

SanjayVas:


stevenwarejones (changes requested):
> Reviewed 1 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXqD0ld8vzFWg37s2lT:b-qfuzjs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*
> ```proto
>         [(google.api.resource_reference).type =
>              "halo.wfanet.org/MeasurementConsumer"];
>     // Matches against the parent `DataProvider`.
> ```

are these comments backwards? eg... if I have a MeasurementConsumer, I filter my results by DataProvider and vice sersa?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 139 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqDC60A-AhizSoHwyP:-NXqDC60A-AhizSoHwyQ:b-pz3evx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L139)):*
> ```proto
>              "halo.wfanet.org/MeasurementConsumer"];
>     // Matches against the parent `DataProvider`.
>     repeated string data_providers = 6
> ```

should these be a oneof? Both data_providers and model_providers shouldn't be specified

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXqDtr-9Pq-MZbGQ_PP:b-23t1xn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*

are these comments backwards? eg... if I have a MeasurementConsumer, I filter my results by DataProvider and vice sersa?
</blockquote></details>

Not sure what you mean. The `measurement_consumers` filter field has a comment stating it matches against the `measurement_consumer` field in the resource message. The `data_providers` filter field has a comment stating it matches against the parent `DataProvider` of the resource.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 139 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqDC60A-AhizSoHwyP:-NXqDWSBDzIZ28_ToOpk:bc1v06x) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L139)):*

should these be a oneof? Both data_providers and model_providers shouldn't be specified
</blockquote></details>

That's only because of our current auth scheme, where the only principals that can call this method are MeasurementConsumers and DataProviders and that each can only list their own child EventGroups. If we happened to allow some admin principal type that can list across all MCs and EDPs, then that caller could theoretically make use of both filters.

I don't think we want to be that prescriptive in filter messages. You can filter however you want. e.g. if the caller is a DataProvider and is therefore using a DataProvider as their request `parent`, then including the same DataProvider in your filter doesn't change the results. Including a different DataProvider in your filter means there will be no results. Neither of this is an error, just not particularly useful.

SanjayVas:
For #144

stevenwarejones (changes requested):
> Reviewed 1 of 2 files at r3, 8 of 8 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXupC1u8pEkz9qZlycn:b-yzp0hb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*

Not sure what you mean. The `measurement_consumers` filter field has a comment stating it matches against the `measurement_consumer` field in the resource message. The `data_providers` filter field has a comment stating it matches against the parent `DataProvider` of the resource.
</blockquote></details>

The comment reads to me that it matches against the parent DataProvider in addition to the provide list of data_providers. In actuality, it filters against the parent MeasurementConsumer data_providers is populated.

I think both these comments should be combined together.

`The Filter filters first by the parent resource and then by the subset that matches the filter fields.`
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 139 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqDC60A-AhizSoHwyP:-NXuptPP7Ahe0phrTYn-:bbvhh9q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L139)):*

That's only because of our current auth scheme, where the only principals that can call this method are MeasurementConsumers and DataProviders and that each can only list their own child EventGroups. If we happened to allow some admin principal type that can list across all MCs and EDPs, then that caller could theoretically make use of both filters.

I don't think we want to be that prescriptive in filter messages. You can filter however you want. e.g. if the caller is a DataProvider and is therefore using a DataProvider as their request `parent`, then including the same DataProvider in your filter doesn't change the results. Including a different DataProvider in your filter means there will be no results. Neither of this is an error, just not particularly useful.
</blockquote></details>

okay. this makes sense if we have bette commenting. see my other comment.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXv3idD-3Xt79J_prdV:btv4sb2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*

The comment reads to me that it matches against the parent DataProvider in addition to the provide list of data_providers. In actuality, it filters against the parent MeasurementConsumer data_providers is populated.

I think both these comments should be combined together.

`The Filter filters first by the parent resource and then by the subset that matches the filter fields.`
</blockquote></details>

A `filter` field in a List method always filters against the resource. Each comment is indicating which part of resource that the subfield is matching. The `measurement_consumers` filter field is matching against the `measurement_consumer` field of the resource. The `data_providers` field is matching against the parent `DataProvider` of the resource (which is an inherent part of the resource name).

This is just a structured version of what would otherwise be supported by [AIP-160](https://google.aip.dev/160). e.g.

```textproto
filter {
  measurement_consumers: "measurementConsumers/abc123"
  measurement_consumers: "measurementConsumers/xyz456"
}
```

Translates to `measurement_consumer =  "measurementConsumers/abc123" OR measurement_consumer = "measurementConsumers/xyz456"`

The `data_providers` field in the `Filter` message translates slightly differently as the EventGroup resource message doesn't have a `data_provider` field, and that is instead indicated by the prefix of the resource name.

```textproto
filter {
  data_providers: "dataProvider/abc123"
}
```

Translates to `name = "dataProviders/abc123/*`

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXv6cYY6v0t0nexJ1rz:ba2m75q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*

A `filter` field in a List method always filters against the resource. Each comment is indicating which part of resource that the subfield is matching. The `measurement_consumers` filter field is matching against the `measurement_consumer` field of the resource. The `data_providers` field is matching against the parent `DataProvider` of the resource (which is an inherent part of the resource name).

This is just a structured version of what would otherwise be supported by [AIP-160](https://google.aip.dev/160). e.g.

```textproto
filter {
  measurement_consumers: "measurementConsumers/abc123"
  measurement_consumers: "measurementConsumers/xyz456"
}
```

Translates to `measurement_consumer =  "measurementConsumers/abc123" OR measurement_consumer = "measurementConsumers/xyz456"`

The `data_providers` field in the `Filter` message translates slightly differently as the EventGroup resource message doesn't have a `data_provider` field, and that is instead indicated by the prefix of the resource name.

```textproto
filter {
  data_providers: "dataProvider/abc123"
}
```

Translates to `name = "dataProviders/abc123/*`
</blockquote></details>

Clarifications just in case it's unclear:
* The filters will apply to the resource collection indicated by the `parent` field in the request. If the `parent` is `measurementConsumers/abc123`, then it means the collection you're referring to is EventGroups belonging to that MC. Further filtering by the `measurement_consumer` field is therefore unnecessary.
* The canonical parent of the `eventGroups` resource collection is `DataProvider`. This is evidenced by the fact that he the canonical format of the EventGroup resource name is `dataProviders/{data_provider}/eventGroups/{event_group}`. That is the format that the service will always return regardless of what format is used for the `parent` of the ListEventGroups request.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXvijf8986spwvqUo3Y:bb7ofbo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*

Clarifications just in case it's unclear:
* The filters will apply to the resource collection indicated by the `parent` field in the request. If the `parent` is `measurementConsumers/abc123`, then it means the collection you're referring to is EventGroups belonging to that MC. Further filtering by the `measurement_consumer` field is therefore unnecessary.
* The canonical parent of the `eventGroups` resource collection is `DataProvider`. This is evidenced by the fact that he the canonical format of the EventGroup resource name is `dataProviders/{data_provider}/eventGroups/{event_group}`. That is the format that the service will always return regardless of what format is used for the `parent` of the ListEventGroups request.
</blockquote></details>

^The filters will apply to the resource collection indicated by the `parent` field in the request. If the `parent` is `measurementConsumers/abc123`, then it means the collection you're referring to is EventGroups belonging to that MC. Further filtering by the `measurement_consumer` field is therefore unnecessary.

I agree with everything you said. I think its confusing potentially to an adopter that if you are a measurementConsumer then there is no point to filling out the measurement_consumer field. I think the comments should reflect that.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 138 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/148#-NXqD0ld8vzFWg37s2lS:-NXwCTpq0H0_2hgYKvII:b-kziu7m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6fc6ecc98ba92be12e25c37c45f15a44bb407dab/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L138)):*

^The filters will apply to the resource collection indicated by the `parent` field in the request. If the `parent` is `measurementConsumers/abc123`, then it means the collection you're referring to is EventGroups belonging to that MC. Further filtering by the `measurement_consumer` field is therefore unnecessary.

I agree with everything you said. I think its confusing potentially to an adopter that if you are a measurementConsumer then there is no point to filling out the measurement_consumer field. I think the comments should reflect that.
</blockquote></details>

It's unnecessary but also doesn't do any harm.

Again, this is just the filtering feature that the AIPs describe but with a more limited structure instead of supporting the entire expression language. I want to avoid having the filter documentation be aware of specific use cases. It shouldn't describe anything more than the resulting predicate.

PR #164, opened 2023-07-28
add population messages

stevenwarejones:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)

SanjayVas (changes requested):
> Reviewed 8 of 8 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nah5SOMECEC-PPB1FG0:b-lgg31n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*
> ```proto
>   message PopulationBlob {
>     // The path the model blob can be downloaded from.
>     string population_blob_path = 1 [
> ```

The blob path refers to the path inside of a given bucket. I think we also need to include bucket and provider fields. Alternatively we can use blob URIs instead of blob paths, e.g. `gs://bucket/path/to/blob` or `s3://bucket/path/to/blob`.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-Nah4pRd8KMeMSjvfXWI:b-t1jc9m) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*
> ```proto
>   }
> 
>   // The blob for the population. Optional.
> ```

1. The API linter will probably complain about this being in the comment. Currently anything not marked as `REQUIRED` is assumed to be optional.
2. I thought we discussed this actually being required? i.e. the flow is "upload your blob, then tell the API where you put it".



_Suggestion:_
```proto
 Optional
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition.proto` line 228 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah7EZ_0gBRwjjDDuJG:-Nah7EZ_0gBRwjjDDuJH:bzgomsl) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L228)):*
> ```proto
> 
>   // Resource name of the ModelLine for this Requisition.
>   string model_line = 14 [
> ```

Where does this come from? If it's on the Requisition only (e.g. not explicitly denormalized from the Measurement), then it implies the Kingdom chooses it. Is that the intent?

@kungfucraig 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah8AO8AXyEkvL-Lodi:-Nah8AO8AXyEkvL-Lodj:bwovezb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L83)):*
> ```proto
>   ];
> 
>   message Events {
> ```

FYI for the implementation, the PR that updates the main repo to this version will need to update at least the following pieces of logic:

* Reporting v1 and v2 implementations
* EDP simulator
* MC simulator (correctness test harness)
* MeasurementSystem CLI tool

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NahTcPj0225DaZjB7F3:bhppsz9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

The blob path refers to the path inside of a given bucket. I think we also need to include bucket and provider fields. Alternatively we can use blob URIs instead of blob paths, e.g. `gs://bucket/path/to/blob` or `s3://bucket/path/to/blob`.
</blockquote></details>

this isn't what we have done elsewhere for blob path. For example the vid model path is not a uri. I think the current design would just assume an agreed upon provider and bucket out-of-band. I think if we fix that here, we need to fix it in the other places as well.

https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c76b38d543854ce47d5be25d2bfe4cbf6d6a9ffe/src/main/proto/wfa/measurement/api/v2alpha/model_shard.proto#L51
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition.proto` line 228 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah7EZ_0gBRwjjDDuJG:-NahTY0U4OTzzelYHEsq:b-z2cpg0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L228)):*

Where does this come from? If it's on the Requisition only (e.g. not explicitly denormalized from the Measurement), then it implies the Kingdom chooses it. Is that the intent?

@kungfucraig 
</blockquote></details>

this should be denormalized from the Measurement. I will add it to the Measurement (missed that)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 5 of 8 files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NahVb2UBtTXWioKg43I:b5x7q2g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

1. The API linter will probably complain about this being in the comment. Currently anything not marked as `REQUIRED` is assumed to be optional.
2. I thought we discussed this actually being required? i.e. the flow is "upload your blob, then tell the API where you put it".
</blockquote></details>

If you are your own population_data_provider, why would you need to provide this to the Kingdom? 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah8AO8AXyEkvL-Lodi:-NahW3sr3ZtQ47RRtoTw:b-b3fasr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L83)):*

FYI for the implementation, the PR that updates the main repo to this version will need to update at least the following pieces of logic:

* Reporting v1 and v2 implementations
* EDP simulator
* MC simulator (correctness test harness)
* MeasurementSystem CLI tool
</blockquote></details>

i can try to have that ready when this is merged so that doesn't block others

SanjayVas (changes requested):
> Reviewed 1 of 3 files at r2.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 6 of 8 files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NahW1mx9W-egRpZC1SX:bqy5war) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

this isn't what we have done elsewhere for blob path. For example the vid model path is not a uri. I think the current design would just assume an agreed upon provider and bucket out-of-band. I think if we fix that here, we need to fix it in the other places as well.

https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c76b38d543854ce47d5be25d2bfe4cbf6d6a9ffe/src/main/proto/wfa/measurement/api/v2alpha/model_shard.proto#L51
</blockquote></details>

The places where we've used blob paths  in the past have been for when the service host owns the blob (e.g. in the internal Duchy API). In those cases, the bucket and provider are part of the configuration for that component.

If the API host isn't the blob owner then the storage configuration cannot be assumed and should be explicit in the API. It's just setting up access to that blob (which could be at a more granular level) that would be handled out-of-band.

So yes, it should be changed elsewhere in the API if the blob is not owned by the service host.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NahXJfK4K343pOeyN1C:b-ht3jg9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

If you are your own population_data_provider, why would you need to provide this to the Kingdom? 
</blockquote></details>

From the design discussion, I thought the decision was that the API and therefore the reference implementation of population requisition fulfillment would be agnostic to who the PDP might be (i.e. whether it's a different entity than the actual survey provider). Therefore the reference implementation should be consuming the API in order to determine the blob path.

Said another way, this provides consistency across types of PDPs.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NahftccDQEkdNIjuXBl:b-whkilf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

The places where we've used blob paths  in the past have been for when the service host owns the blob (e.g. in the internal Duchy API). In those cases, the bucket and provider are part of the configuration for that component.

If the API host isn't the blob owner then the storage configuration cannot be assumed and should be explicit in the API. It's just setting up access to that blob (which could be at a more granular level) that would be handled out-of-band.

So yes, it should be changed elsewhere in the API if the blob is not owned by the service host.
</blockquote></details>

@kungfucraig - do you see any issues with exposing this information even when its not needed by anyone but the Population Data Provider?

Or should it just be a config internal to the Population Data Provider?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 57 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nahigok0RbJyzb3KY9-:-Nahigok0RbJyzb3KY90:bdmepiy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/746475c259568ec17ba68e866ef876955049ce4d/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L57)):*
> ```proto
> 
>   // The blob for the population.
>   PopulationBlob population_blob = 4 [(google.api.field_behavior) = IMMUTABLE];
> ```

@kungfucraig @SanjayVas - I was thinking maybe this blob should be encrypted with the PDP's public key? If so, I'd want to include that certificate here. wdyt?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 57 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nahigok0RbJyzb3KY9-:-Nahm81fC-rXdSX9WYtg:bs4w2x6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/746475c259568ec17ba68e866ef876955049ce4d/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L57)):*

@kungfucraig @SanjayVas - I was thinking maybe this blob should be encrypted with the PDP's public key? If so, I'd want to include that certificate here. wdyt?
</blockquote></details>

If that's the case, we'd also need to note which public key was used (since the PDP is expected to have many keys over time). We should also then note what the format is so you know how to extract the content and signature from the plaintext (we tend to use sign-then-encrypt).

As for whether this *should* always be encrypted with the PDP's public key, If it is, then only the PDP can decrypt it (meaning the future use case of just giving the MP access to the blob wouldn't work). A signature makes sense in that case so an MP can verify the blob came from the PDP. In either case, I assume the PDP owns the blob storage.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition.proto` line 228 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah7EZ_0gBRwjjDDuJG:-NahjlwQ7BK9ee_xY_UP:bd7i6bd) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L228)):*

this should be denormalized from the Measurement. I will add it to the Measurement (missed that)
</blockquote></details>

If the intent is for the MC to specify this, maybe it should go in the signed MeasurementSpec?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 6 of 8 files reviewed, 4 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition.proto` line 228 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah7EZ_0gBRwjjDDuJG:-Nai1q5m6TQS0aXtJd0h:b-d7bhn0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L228)):*

If the intent is for the MC to specify this, maybe it should go in the signed MeasurementSpec?
</blockquote></details>

will do

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nai5yGs0QA1ZMkUIut3:bgpfczq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

@kungfucraig - do you see any issues with exposing this information even when its not needed by anyone but the Population Data Provider?

Or should it just be a config internal to the Population Data Provider?
</blockquote></details>

In the case of model_shard, the resource parent is a DataProvider. It will be uploaded by the ModelProvider and downloaded by the DataProvider. So, we should probably change that to URI too? @Marco-Premier 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-Nai58M271JuVA4_7ZRQ:bihpv4c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

From the design discussion, I thought the decision was that the API and therefore the reference implementation of population requisition fulfillment would be agnostic to who the PDP might be (i.e. whether it's a different entity than the actual survey provider). Therefore the reference implementation should be consuming the API in order to determine the blob path.

Said another way, this provides consistency across types of PDPs.
</blockquote></details>

moving this comment to the correct thread....
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition.proto` line 228 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah7EZ_0gBRwjjDDuJG:-Nai4xuj5wljCzJ7Oq5C:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L228)):*

will do
</blockquote></details>

Done.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-Nai6XOJ0GtjbXVibxJI:b1626oa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

moving this comment to the correct thread....
</blockquote></details>

[@kungfucraig](https://github.com/kungfucraig) - do you see any issues with exposing this information even when its not needed by anyone but the Population Data Provider?

Or should it just be a config internal to the Population Data Provider?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 3 of 8 files reviewed, 4 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 57 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nahigok0RbJyzb3KY9-:-Nai6kT32cj2fMmlz50Y:b-nyx07s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/746475c259568ec17ba68e866ef876955049ce4d/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L57)):*

If that's the case, we'd also need to note which public key was used (since the PDP is expected to have many keys over time). We should also then note what the format is so you know how to extract the content and signature from the plaintext (we tend to use sign-then-encrypt).

As for whether this *should* always be encrypted with the PDP's public key, If it is, then only the PDP can decrypt it (meaning the future use case of just giving the MP access to the blob wouldn't work). A signature makes sense in that case so an MP can verify the blob came from the PDP. In either case, I assume the PDP owns the blob storage.
</blockquote></details>

yeah - maybe we should wait on this one -- for those future use cases, if its encrypted, separate populations would need to be created for each party.

SanjayVas (commented):
> Reviewed 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 2 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, and @Marco-Premier)

kungfucraig (changes requested):
> Reviewed 2 of 8 files at r1, 1 of 3 files at r2, 4 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 161 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nali_XF1LIbzdOv8lsl:-Nali_XF1LIbzdOv8lsm:bk3s3u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/682e2f4c23cdd4437eec2c1525bedec3d6bb05ae/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L161)):*
> ```proto
> 
>     // A population result.
>     message Population {
> ```

Once @riemanli merges his PR around the methodologies, will it be worth adding something here? 

@stevenwarejones we discussed a number of different ways to implement the population estimate across model rollouts. It seems worth making that explicit here even if we only support one way in the initial version. 

cc: @SanjayVas 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nalkg3lAuD7XwLVc1da:b8osnhs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

In the case of model_shard, the resource parent is a DataProvider. It will be uploaded by the ModelProvider and downloaded by the DataProvider. So, we should probably change that to URI too? @Marco-Premier 
</blockquote></details>

I think the analogue for Population is Exchange and as far as I can tell the blob path is not stored by the Kingdom at all, which means it's an out of band configuration that is agreed upon between the MP and the EDP. We should do the same thing here. 

Or, if I'm mistaken about the panel exchange, we should still do whatever it does because this is basically facilitating a similar type of relationship.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-Nalm27T68CzMDiTxBQT:bkcxdfz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

[@kungfucraig](https://github.com/kungfucraig) - do you see any issues with exposing this information even when its not needed by anyone but the Population Data Provider?

Or should it just be a config internal to the Population Data Provider?
</blockquote></details>

See above thread.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah8AO8AXyEkvL-Lodi:-NalmPgk0Q_Uf9ZdOhQz:blz3pqa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L83)):*

i can try to have that ready when this is merged so that doesn't block others
</blockquote></details>

Should this generate a release note indicating that EDPs need to migrate off of the event_groups field above?

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 2 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NalzcRy8T6-frHv8l8P:bdh8e9p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*
> See above thread.

I don't think the above thread response really answers the question of whether this field should be unconditionally required.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 2 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NamBbJV2nWJO-pKNGLW:bkk9w7c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

> See above thread.

I don't think the above thread response really answers the question of whether this field should be unconditionally required.
</blockquote></details>

I'm actually suggesting we drop the Population blob from the API altogether and require that anyone who needs it make arrangements out of band to get access to it. This is inline with the panelist exchange as far as I can tell.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 2 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NamEhiUBHEwu3FnF8Ci:brzjjan) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*
> This is inline with the panelist exchange as far as I can tell.

There is a workflow step that references a blob path (`CopyFromPreviousExchangeStep`).

It basically depends on how much of the coordination you want to go through the Kingdom. I don't have a huge concern with the PDP just having to maintain a Population resource to blob mapping internally, but note that granting of access to a blob can be done unilaterally after some initial setup (e.g. setting up the service account). There's no need for a PDP and MP to actually communicate again for each new Population resource if the resource has a blob ref.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 161 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nali_XF1LIbzdOv8lsl:-NamLxHw4LVWSDhzeRUV:b-uktzkh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/682e2f4c23cdd4437eec2c1525bedec3d6bb05ae/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L161)):*

Once @riemanli merges his PR around the methodologies, will it be worth adding something here? 

@stevenwarejones we discussed a number of different ways to implement the population estimate across model rollouts. It seems worth making that explicit here even if we only support one way in the initial version. 

cc: @SanjayVas 
</blockquote></details>

I think we should add methodology fields later. I don't think we should add anything now.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 112 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-NamGNqb0KU-3YICUrmz:-NamLrUJBrrQaG0axnIQ:b-7kdpa0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/682e2f4c23cdd4437eec2c1525bedec3d6bb05ae/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L112)):*

From offline discussion with @kungfucraig, we might want to have this not be required initially since it isn't likely to be used for initial trials.
</blockquote></details>

I agree that for Alpha, it shouldn't be required but then we can change it to required for Beta? Wdyt @SanjayVas 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NamIoCw4gl6NE2RkxoX:b-ybcq4j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

I think the analogue for Population is Exchange and as far as I can tell the blob path is not stored by the Kingdom at all, which means it's an out of band configuration that is agreed upon between the MP and the EDP. We should do the same thing here. 

Or, if I'm mistaken about the panel exchange, we should still do whatever it does because this is basically facilitating a similar type of relationship.
</blockquote></details>

for panel exchange, its a path based on the ExchangeDateKey using an out-of-band pre-defined storage. We don't really have an equivalent pre-defined resource type like ExchangeDateKey here.... eg the Census Provider needs a way to upload the blob before the Population resource is defined. But if we were to define such a system, steps would be:
1. Delete this field entirely
2. PDP and Census Provider agree on a shared storage out-of-band
3. Census Provider uploads the blob to pre-determined blob path
4. PDP registers Population Resource
5. When PDP gets a population requisition, it already knows that pre-determined blob path

The other way to go about it is to convert this field to uri
1. PDP gives Census Provider blob upload rights to a bucket out-of-band
2. Census Provider uploads the blob to a new blob uri out-of-band
3. Census Provider gives blob uri to PDP out-of-band
3. PDP registers Population Resource with blob uri
4. When PDP gets a population requisition, it determines blob uri from the Population resource and downloads it
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NamKn2j8Hc1CS0zghH9:b-gtxwes) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

> This is inline with the panelist exchange as far as I can tell.

There is a workflow step that references a blob path (`CopyFromPreviousExchangeStep`).

It basically depends on how much of the coordination you want to go through the Kingdom. I don't have a huge concern with the PDP just having to maintain a Population resource to blob mapping internally, but note that granting of access to a blob can be done unilaterally after some initial setup (e.g. setting up the service account). There's no need for a PDP and MP to actually communicate again for each new Population resource if the resource has a blob ref.
</blockquote></details>

What would be the steps for the PDP to create this mapping?
1. Get the blob path/uri
2. Create the Population resource
3. Map resource to uri?

If that's the flow, its undesirable that the PDP has to create the Population resource before they can create the mapping. I guess we could create a single client tool that immediately creates the map (in a db?) once the kingdom returns the resource name but it feels messy to me.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, @riemanli, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NamWG0A6RqJ8PHbLIX4:ba5qe4z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*
> What would be the steps for the PDP to create this mapping?

This depends on your priorities and whether you're generating random blob keys or have some way to derive the blob key from the population data itself. The latter makes it simpler to avoid orphan blobs. For each of the examples below we'll assume you're using the blob key as the request ID for the CreatePopulation request so it can be idempotent.

If you are generating blob keys and priority is avoiding orphan blobs, then the PDP would need to do something like the following:

1. Create a DB entry for the blob ref with a "not uploaded" state.
2. Upload the blob
3. Update the DB entry to indicate that the blob has been uploaded.
4. Create the Population resource
5. Update the DB entry with the Population resource name.

If you are generating blob keys and are not concerned about orphan blobs, then it can be:

1. Upload blob
2. Write DB entry with blob ref
3. Create Population resource
4. Update DB entry with Population resource name

If you can derive the blob key, then it can be:

1. Upload blob
2. Create Population resource
3. Write DB entry with Population resource name

If you're not concerned about orphan blobs *and* the Population resource has a blob ref:

1. Upload blob
2. Create Population resource

Notice how in the last case the PDP doesn't need to have a local DB.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NamlQAh8H9bZbcVDbPS:bnhgdlj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

> What would be the steps for the PDP to create this mapping?

This depends on your priorities and whether you're generating random blob keys or have some way to derive the blob key from the population data itself. The latter makes it simpler to avoid orphan blobs. For each of the examples below we'll assume you're using the blob key as the request ID for the CreatePopulation request so it can be idempotent.

If you are generating blob keys and priority is avoiding orphan blobs, then the PDP would need to do something like the following:

1. Create a DB entry for the blob ref with a "not uploaded" state.
2. Upload the blob
3. Update the DB entry to indicate that the blob has been uploaded.
4. Create the Population resource
5. Update the DB entry with the Population resource name.

If you are generating blob keys and are not concerned about orphan blobs, then it can be:

1. Upload blob
2. Write DB entry with blob ref
3. Create Population resource
4. Update DB entry with Population resource name

If you can derive the blob key, then it can be:

1. Upload blob
2. Create Population resource
3. Write DB entry with Population resource name

If you're not concerned about orphan blobs *and* the Population resource has a blob ref:

1. Upload blob
2. Create Population resource

Notice how in the last case the PDP doesn't need to have a local DB.

</blockquote></details>

@kungfucraig - wdyt about that last option. I agree with the simplicity of not needing a local DB. On the other hand, it does imply some trust in the kingdom (which is hard to avoid without redesigning other portions as well).

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, @riemanli, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 161 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nali_XF1LIbzdOv8lsl:-NamkZl92EJGBh5Xnc9Y:b-uzjxxt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/682e2f4c23cdd4437eec2c1525bedec3d6bb05ae/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L161)):*

I think we should add methodology fields later. I don't think we should add anything now.
</blockquote></details>

Sounds good. Just wanted to make sure it's on our radar.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NammQaI4JduiQR6Hquh:b-grkig3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

@kungfucraig - wdyt about that last option. I agree with the simplicity of not needing a local DB. On the other hand, it does imply some trust in the kingdom (which is hard to avoid without redesigning other portions as well).
</blockquote></details>

So then the blob path is stored in the Population resource but relative to a bucket agreed to and permissioned out of band?

Like this: 
1. PDP determines blob path relative to predetermined bucket
2. PDP uploads blob
3. PDP creates Population resource with the blob path

If so, sgtm

riemanli (commented):
> Reviewed 2 of 8 files at r1, 5 of 5 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, @SanjayVas, and @stevenwarejones)

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 2 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 112 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-NamGNqb0KU-3YICUrmz:-NarQgQH5vE4BNDMOhPr:b-hdz91) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/682e2f4c23cdd4437eec2c1525bedec3d6bb05ae/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L112)):*

I agree that for Alpha, it shouldn't be required but then we can change it to required for Beta? Wdyt @SanjayVas 
</blockquote></details>

Yeah, that sounds good. I think implementation-wise we'd have it set by Reporting v2 but not by v1.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NarRhgOEeb5s_6rzO6I:bz74ofa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

for panel exchange, its a path based on the ExchangeDateKey using an out-of-band pre-defined storage. We don't really have an equivalent pre-defined resource type like ExchangeDateKey here.... eg the Census Provider needs a way to upload the blob before the Population resource is defined. But if we were to define such a system, steps would be:
1. Delete this field entirely
2. PDP and Census Provider agree on a shared storage out-of-band
3. Census Provider uploads the blob to pre-determined blob path
4. PDP registers Population Resource
5. When PDP gets a population requisition, it already knows that pre-determined blob path

The other way to go about it is to convert this field to uri
1. PDP gives Census Provider blob upload rights to a bucket out-of-band
2. Census Provider uploads the blob to a new blob uri out-of-band
3. Census Provider gives blob uri to PDP out-of-band
3. PDP registers Population Resource with blob uri
4. When PDP gets a population requisition, it determines blob uri from the Population resource and downloads it
</blockquote></details>

The assumption is that the PDP owns the blob. This Halo system need not care about the relationship between the PDP and the census provider, so how the data gets to the PDP need not be considered as part of this API design. It's the PDP's responsibility to make sure the blob gets into the right place at the right time, which could entail copying it from another location.

What kind of out-of-band communication the PDP and MP would need to have if this isn't a full URI depends on how things are configured. e.g. if each PDP always uses a fixed single bucket then the MP just needs a static config mapping PDP to bucket on creation. If any PDP might not always use the same static bucket then the amount of communication and how the PDP and MP need to keep track of the mappings becomes more complex.

Our reference fullfiller for the PDP would be configured to use a single bucket, but there's no guarantee that other PDPs will do the same (unless we document that requirement somewhere).
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-Namdfcm0Nc8eCYMT2xc:beou9fo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*
> So then the blob path is stored in the Population resource but relative to a bucket agreed to and permissioned out of band?

The comment threads got a little conflated. That part is being discussed in the other comment thread. This is just whether some blob ref should be a required part of the resource independently of whether that ref is just a blob key vs. a URI.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 2 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NarVEVhBsq0Tip2Eo2d:b-r2cprs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

The assumption is that the PDP owns the blob. This Halo system need not care about the relationship between the PDP and the census provider, so how the data gets to the PDP need not be considered as part of this API design. It's the PDP's responsibility to make sure the blob gets into the right place at the right time, which could entail copying it from another location.

What kind of out-of-band communication the PDP and MP would need to have if this isn't a full URI depends on how things are configured. e.g. if each PDP always uses a fixed single bucket then the MP just needs a static config mapping PDP to bucket on creation. If any PDP might not always use the same static bucket then the amount of communication and how the PDP and MP need to keep track of the mappings becomes more complex.

Our reference fullfiller for the PDP would be configured to use a single bucket, but there's no guarantee that other PDPs will do the same (unless we document that requirement somewhere).
</blockquote></details>

Note that there's still a benefit to having the blob ref here even if it's just the key/path and the bucket is communicated out-of-band (see the other comment thread).

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nas7Xyh1ORWrZjx6VHZ:bd197gt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

Note that there's still a benefit to having the blob ref here even if it's just the key/path and the bucket is communicated out-of-band (see the other comment thread).
</blockquote></details>

@stevenwarejones Can you update the comment on the field? Then I think we're good to go.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 6 of 8 files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 112 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-NamGNqb0KU-3YICUrmz:-NavRWu-EpPPqrklWpk9:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/682e2f4c23cdd4437eec2c1525bedec3d6bb05ae/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L112)):*

Yeah, that sounds good. I think implementation-wise we'd have it set by Reporting v2 but not by v1.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NavRaQj7cL5KkqgxN2P:b-jcok7n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

@stevenwarejones Can you update the comment on the field? Then I think we're good to go.
</blockquote></details>

Changed this to model_blob_ref and added more comments. I like keeping it generic. It could be based on a rotating bucket for example.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 55 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah4pRd8KMeMSjvfXWH:-NavRlxr1RAngsnFu7cj:b8pikiw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L55)):*

> So then the blob path is stored in the Population resource but relative to a bucket agreed to and permissioned out of band?

The comment threads got a little conflated. That part is being discussed in the other comment thread. This is just whether some blob ref should be a required part of the resource independently of whether that ref is just a blob key vs. a URI.
</blockquote></details>

Based on discusssion changed this to REQUIRED

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NawX41A2tEXM_ahc9SE:bwfsdlw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

Changed this to model_blob_ref and added more comments. I like keeping it generic. It could be based on a rotating bucket for example.
</blockquote></details>

Sorry, what we decided on was just that we should have a blob ref (path or URI) in this resource. I was using the term "blob ref" just because we hadn't decided yet on which one. The API **should** be opinionated about which one. Ideally it should *always* be one or the other. If we *really* want to support either (which I don't think we should), it should be a oneof so it's explicit.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nb0-DSt0aU1NDqsK4mo:b1cdhoh) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

Sorry, what we decided on was just that we should have a blob ref (path or URI) in this resource. I was using the term "blob ref" just because we hadn't decided yet on which one. The API **should** be opinionated about which one. Ideally it should *always* be one or the other. If we *really* want to support either (which I don't think we should), it should be a oneof so it's explicit.
</blockquote></details>

i had considered a oneof as well. Any opinions why we shouldn't support both?

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nb01eKfCK-VC3PSxHXm:b-mt52mw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

i had considered a oneof as well. Any opinions why we shouldn't support both?
</blockquote></details>

To clarify, I think our reference implementation will only support one, but I think supporting multiple (including reference_id) has value in different use cases.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah8AO8AXyEkvL-Lodi:-Nb01oSvEiKaQvVF4hLw:brb5gt6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L83)):*

Should this generate a release note indicating that EDPs need to migrate off of the event_groups field above?
</blockquote></details>

That could come after we get cross-media-measurement to adopt these API changes.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nb13KsW8fpWzS0_9Q0F:b-t2atl7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

To clarify, I think our reference implementation will only support one, but I think supporting multiple (including reference_id) has value in different use cases.
</blockquote></details>

Simplicity (avoid having choices unless we need them). 

Right now we should follow the YAGNI principle and pick one. If we need to add options later, wrapping non-repeated fields in a `oneof` is a backwards-compatible change.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 7 of 8 files reviewed, 1 unresolved discussion (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nb1ItFm5VcX1DRtPn8A:baoolax) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

Simplicity (avoid having choices unless we need them). 

Right now we should follow the YAGNI principle and pick one. If we need to add options later, wrapping non-repeated fields in a `oneof` is a backwards-compatible change.
</blockquote></details>

lets go with uri then. I think its the most flexible to start with since it covers the path use case implicitly.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @EvgSkv, @jojijac0b, @kungfucraig, and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-Nb1M0mWAZYB5h7S8k2N:brnhlq3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

lets go with uri then. I think its the most flexible to start with since it covers the path use case implicitly.
</blockquote></details>

Works for me since it's what I asked for initially, but I think @kungfucraig wanted to start with implicit bucket.

kungfucraig (approved):
> Reviewed 1 of 5 files at r3, 1 of 2 files at r4, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @EvgSkv, @jojijac0b, and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/population.proto` line 49 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nah5SOMECEC-PPB1FG-:-NbKh9w4DaPqSNWZwSk7:b-kfkrvs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ac8651b1391344640e0446a547463037f2793aa8/src/main/proto/wfa/measurement/api/v2alpha/population.proto#L49)):*

Works for me since it's what I asked for initially, but I think @kungfucraig wanted to start with implicit bucket.
</blockquote></details>

URI is fine. :)

stevenwarejones (commented):
@SanjayVas - can you reapprove the latest change?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-:-NbX9gVV5PzNektr8GWE:b4uzfic)* status: 8 of 9 files reviewed, all discussions resolved (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, and @SanjayVas)

SanjayVas (changes requested):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: all files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel` line 140 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nb_f_k44xNbMuQReBT2:-Nb_f_k55t7lGTq60zJ6:b-a8t6ay) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2194706d760df3ef03e44f1c393253a5cf121369/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel#L140)):*
> ```Starlark
>         "@com_google_googleapis//google/api:field_behavior_proto",
>         "@com_google_googleapis//google/api:resource_proto",
>         "@com_google_protobuf//:any_proto",
> ```

I don't see all of these dependencies used by `event_template.proto`



_Suggestion:_
```Starlark
        "@com_google_googleapis//google/api:field_behavior_proto",
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 98 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nb_gC8s95BA9lMf9opo:-Nb_gC8s95BA9lMf9opp:b-5q9jr3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2194706d760df3ef03e44f1c393253a5cf121369/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L98)):*
> ```proto
>     // will be matched. Required if measurement type is population.
>     EventFilter filter = 1;
>     // Time interval over which the event data should be collected.
> ```

I don't think this comment is correct, as there is no event data for a population measurement.



_Code quote:_
```proto
event data should be collected
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 103 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nb_gMcd4fL_YmaqzAJ1:-Nb_gMcd4fL_YmaqzAJ2:b4yqaif) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2194706d760df3ef03e44f1c393253a5cf121369/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L103)):*
> ```proto
>   }
> 
>   oneof requisition_type {
> ```

Add a comment for this `oneof`. Indicate that it's required. There's no syntax for applying the field behavior annotation to a `oneof`, so we have to use a comment to indicate that one of the fields in the `oneof` must be specified.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 7 of 10 files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel` line 140 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nb_f_k44xNbMuQReBT2:-Nb_paqvBiiE8EnXssum:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2194706d760df3ef03e44f1c393253a5cf121369/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel#L140)):*

I don't see all of these dependencies used by `event_template.proto`
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 98 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nb_gC8s95BA9lMf9opo:-Nb_pdJJ82DwjRq5RwdI:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2194706d760df3ef03e44f1c393253a5cf121369/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L98)):*

I don't think this comment is correct, as there is no event data for a population measurement.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto` line 103 at r7](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164#-Nb_gMcd4fL_YmaqzAJ1:-Nb_peWL-1GgmTKvcnxM:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2194706d760df3ef03e44f1c393253a5cf121369/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto#L103)):*

Add a comment for this `oneof`. Indicate that it's required. There's no syntax for applying the field behavior annotation to a `oneof`, so we have to use a comment to indicate that one of the fields in the `oneof` must be specified.
</blockquote></details>

Done.

kungfucraig (approved):
> Reviewed 1 of 1 files at r7, 2 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: 9 of 10 files reviewed, 3 unresolved discussions (waiting on @EvgSkv, @jojijac0b, @Marco-Premier, @riemanli, and @SanjayVas)

SanjayVas (approved):
> Reviewed 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/164)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @EvgSkv, @jojijac0b, and @Marco-Premier)

PR #165, opened 2023-08-21
Clarify specification of maximum frequency.

### LINKED ISSUE #159: Duration MeasurementSpecs should not have max_frequency_per_user
This field implies that the `DataProvider` should cap the maximum number of events considered per user. We use this field for DP sensitivity in `Impression` Measurements, but we do not want it to be used for `Duration` measurements. The name is also misleading for `ReachAndFrequency`, where it is clearer to describe it as the maximum frequency revealed in the final histogram.

Note that much of our code does not even support this field for those Measurement types already.

- [x] #165 
- [x] https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1183

### LINKED ISSUE #160: maximum_frequency should be specified in MeasurementSpec not protocol config
This is not protocol-specific. We also want it specified consistently across both MPC-based and direct R/F measurements.

Related to #159

- [x] #165 
- [x] https://github.com/world-federation-of-advertisers/cross-media-measurement/pull/1183

https://rally1.rallydev.com/#/?detail=/userstory/720468381001&fdp=true

SanjayVas:
* Deprecate the `maximum_frequency` field in `ProtocolConfig.LiquidLegionsV2` so it is no longer specified in two places.
* Rename `maximum_frequency_per_user` in `MeasurementSpec.ReachAndFrequency` to `maximum_frequency`, clarifying the requirements.
* Drop the unsupported `maximum_frequency_per_user` field in `MeasurementSpec.Duration`.

See #159 and #160

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -63,8 +63,13 @@ message MeasurementSpec {
     // Differential privacy parameters for frequency.
     DifferentialPrivacyParams frequency_privacy_params = 2;
 
-    // Maximum frequency per user that will be included in this measurement.
-    int32 maximum_frequency_per_user = 3;
+    // The maximum frequency to reveal in the distribution.
+    //
+    // This is required in new `CreateMeasurement` requests, but may not be
+    // specified for legacy `Measurement`s. In the case where it is not
+    // specified, it must instead be read from the corresponding field in
+    // `ProtocolConfig`.
+    int32 maximum_frequency = 3;
   }
 
   // Parameters for an impression measurement.
@@ -78,15 +83,14 @@ message MeasurementSpec {
 
   // Parameters for a duration measurement.
   message Duration {
+    reserved 3;
+
     // Differential privacy parameters.
     DifferentialPrivacyParams privacy_params = 1;
 
     // Maximum watch duration in seconds per user that will be included in this
     // measurement.
     int32 maximum_watch_duration_per_user = 2;
-
-    // Maximum frequency per user that will be included in this measurement.
-    int32 maximum_frequency_per_user = 3;
   }
 
   // Parameters for a population measurement.
--- a/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
@@ -93,7 +93,10 @@ message ProtocolConfig {
     int32 elliptic_curve_id = 3 [(google.api.field_behavior) = REQUIRED];
 
     // The maximum frequency to reveal in the histogram.
-    int32 maximum_frequency = 4;
+    //
+    // Deprecated: Specified in MeasurementSpec instead except for legacy
+    // Measurements.
+    int32 maximum_frequency = 4 [deprecated = true];
 
     // The mechanism to generate noise during computation.
     //


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)

jiayu-google (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqq3uCAoNOK0q-oVs:-NccYLaU2_xGocK25Rx8:b-5olw5i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/85fa94864b9317c43b546e6c5fde5478ba77a0cf/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L67)):*
> Is this new field meant to indicate the same thing as the previous field?

No. Maximum frequency refers to the maximum value revealed in the final histogram, where values above that frequency get attributed to the max frequency bucket. Maximum frequency *per user* instead acts as an event filter where it caps the number of events that a single user contributes to the computation.

The latter isn't actually useful or practical for anything other then impression measurements.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 89 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqs6A6qBfson220lO:-NccY_YrEW-x1PUSEofL:brwsuwe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/85fa94864b9317c43b546e6c5fde5478ba77a0cf/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L89)):*

Are there attacks on privacy budget that maximum_watch_duration_per_user don't handle but maximum_frequency_per_user does?
</blockquote></details>

It's not a question of attacks but rather the usefulness/practicality/feasibility. Basically this field was added to this measurement type in error.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqq3uCAoNOK0q-oVs:-Nccl8ji8gKPmfmEC1eP:b-5131b1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L67)):*

> Is this new field meant to indicate the same thing as the previous field?

No. Maximum frequency refers to the maximum value revealed in the final histogram, where values above that frequency get attributed to the max frequency bucket. Maximum frequency *per user* instead acts as an event filter where it caps the number of events that a single user contributes to the computation.

The latter isn't actually useful or practical for anything other then impression measurements.
</blockquote></details>

it definitely has use here as well. One can compute impression count from a histogram, for example.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqq3uCAoNOK0q-oVs:-Ncclg3r6KS9Ficq315m:bee7haa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L67)):*

it definitely has use here as well. One can compute impression count from a histogram, for example.
</blockquote></details>

I'm not certain of your meaning. max_frequency_per_user was only added for differential privacy purposes w.r.t. direct impression measurements. It was added to these other measurement types in error. We have no plans to support it as we don't need it for DP. See #159.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqq3uCAoNOK0q-oVs:-Ncds87M8NmPLVwXmBFW:bjgveuz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L67)):*
> > Is this new field meant to indicate the same thing as the previous field? So, just a rename?
>
> No. Maximum frequency refers to the maximum value revealed in the final histogram, where values above that frequency get attributed to the max frequency bucket. Maximum frequency _per user_ instead acts as an event filter where it caps the number of events that a single user contributes to the computation.

Turns out I didn't realize after talking to Craig that this will have the same effect w.r.t. a frequency histogram, it just implies something different about how you implement it. That said, still going with adding a new field instead of renaming since MeasurementSpecs are serialized and persisted. Having it be a separate field means we can tell the difference between a legacy Measurement and a new one.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @jiayu-google, @kungfucraig, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqq3uCAoNOK0q-oVs:-NcgznxC3TH_42BVzLrY:b-ixo5n1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L67)):*

> > Is this new field meant to indicate the same thing as the previous field? So, just a rename?
>
> No. Maximum frequency refers to the maximum value revealed in the final histogram, where values above that frequency get attributed to the max frequency bucket. Maximum frequency _per user_ instead acts as an event filter where it caps the number of events that a single user contributes to the computation.

Turns out I didn't realize after talking to Craig that this will have the same effect w.r.t. a frequency histogram, it just implies something different about how you implement it. That said, still going with adding a new field instead of renaming since MeasurementSpecs are serialized and persisted. Having it be a separate field means we can tell the difference between a legacy Measurement and a new one.
</blockquote></details>

yeah - these two fields have the same effect. I hadn't realized what you had meant by maximum_frequency. 

Its the duration field change that will actually change the privacy behavior.

In regards to this, why the name change if the behavior is the same?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 89 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqs6A6qBfson220lO:-Nch-FoJ8HnzVPWzW239:b-vpy3iq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L89)):*

It's not a question of attacks but rather the usefulness/practicality/feasibility. Basically this field was added to this measurement type in error.
</blockquote></details>

@jiayu-google @kungfucraig - this will change the privacy behavior, correct?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: 1 of 2 files reviewed, 2 unresolved discussions (waiting on @jiayu-google, @kungfucraig, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 67 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqq3uCAoNOK0q-oVs:-Ncx29z17hDqdycpXTz9:b-3r7n0p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L67)):*
> In regards to this, why the name change if the behavior is the same?

I updated the issue and PR description to clarify. Even though the end result is the same, it's clearer to describe this as the maximum frequency revealed in the distribution. This makes it more consistent with how the result is described.

I also switched this to being a rename rather than a new field since the meaning hasn't actually changed and it turns out we were actually using this field in the LLv2 implementation (it was using the minimum of this value and the value specified in the protocol config).

kungfucraig (approved):
> Reviewed 1 of 2 files at r2, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @jiayu-google and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 89 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqs6A6qBfson220lO:-Nd2JpCXBFhKzhdIoAv9:b7j9nz8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L89)):*

@jiayu-google @kungfucraig - this will change the privacy behavior, correct?
</blockquote></details>

I'm going to defer to @jiayu-google on impact on privacy. Intuitively max duration and max frequency should be correlated.

jiayu-google (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 89 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165#-Ncbqs6A6qBfson220lO:-Nd2KqpI36oD5BrD01rK:b-9wqdyq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/11b0dbc67b8a47c9fb4319f7c5ebea76d04dd6cc/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L89)):*

I'm going to defer to @jiayu-google on impact on privacy. Intuitively max duration and max frequency should be correlated.
</blockquote></details>

Sorry for replying late, Steven! Good question.

Removing the frequency cap in duration measurement does not break different privacy.
(a) By math definition, the DP of duration measurement only depends on the duration cap, but not the frequency cap.
(b) I can't deny the possibility that the attacking behavior changes when we exclude the frequency cap. Yet DP controls the max info an attacker can gain, regardless of the attacking behavior. I guess that's one reason why DP became a golden standard --- it allows us to not worry about the attacking behavior.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

stevenwarejones (approved):
> Reviewed 1 of 1 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/165)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #175, opened 2023-09-25
Add Data Availability Support

tristanvuong2021:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)

kungfucraig (approved):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfCvdKD2MhMH_yoS4aZ:-NfCvdKD2MhMH_yoS4a_:brvmle8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto#L42)):*
> ```proto
> 
>   // Replaces `data_availability_interval` for a `DataProvider` resource.
>   rpc ReplaceDataAvailabilityInterval(ReplaceDataAvailabilityIntervalRequest)
> ```

I don't like that we used "Replace" over "Set" or "Update" as the verb here. But I suppose we should be consistent. @SanjayVas any opinions?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 118 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfCw-CN0G0DRwcimWqi:-NfCw-CN0G0DRwcimWqj:bdftb95) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L118)):*
> ```proto
> 
>   // Interval for when data is available for a `Requisition`. If `end_time` is
> ```

This field is updated with the standard Update method?

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfDl3A64_5WiLzdHQtK:-NfDl3A7721-VrQjHU92:b-sg6oj8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*
> ```proto
>   // Interval for when data is available for a `Requisition`. If `start_time` is
>   // not set, data is not available. If `end_time` is not set, data is available
>   // from `start_time` to now.
> ```

I don't believe this matches the behavior we discussed. In our discussion, we indicated that this field would be something the DataProvider must update whenever data becomes available. Therefore `end_time` must be set if `data_availability_interval` is specified.

Additionally, I assume we don't want Requisitions to immediately start failing between when this rolls out and when `DataProvider`s set this for the first time. Therefore we might want to phrase this as whether data is *guaranteed* to be available. That gives us the flexibility to e.g. require this to be set for Report scheduling but not for Requisition fulfillment.



_Code quote:_
```proto
  // Interval for when data is available for a `Requisition`. If `start_time` is
  // not set, data is not available. If `end_time` is not set, data is available
  // from `start_time` to now.
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfCvdKD2MhMH_yoS4aZ:-NfDhCvd31cusQ0OaN0V:b44lugf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto#L42)):*

I don't like that we used "Replace" over "Set" or "Update" as the verb here. But I suppose we should be consistent. @SanjayVas any opinions?
</blockquote></details>

"Update" has specific meaning as a standard method which would have to conform with [AIP-134](https://google.aip.dev/134) and let you update any field. We've instead chosen to use custom methods for setting specific fields.

In terms of "Set" vs. "Replace", we use both verbs for method names in this API. "Replace" was first used for `ReplaceAccountIdentity` due to the fact that the identity had to have already been set. It's also used in the above `ReplaceDataProviderRequiredDuchies` method, which has no such restriction.

It's probably best to be consistent with the "ReplaceDataProviderRequiredDuchies" method since it's on the same resource. The alternative IMO would be to use "Set" and rename the other method, but that's a backwards-incompatible change.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 121 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfDm1WpEWyXu7DzjyRs:-NfDm1WpEWyXu7DzjyRt:b-roxzfi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L121)):*
> ```proto
>   // not set, data is available from `start_time` to now. If this is not set,
>   // the `data_availability_interval` in the parent `DataProvider` determines
>   // the data availability.
> ```

This does not match what was previously discussed. The key point is that this is *always* constrained by the range in the parent DataProvider, i.e. the data won't actually be available for any time earlier than the start in the parent DataProvider regardless of what the start here is and won't be available for any time on or later than the end in the parent DataProvider.



_Code quote:_
```proto
  // Interval for when data is available for a `Requisition`. If `end_time` is
  // not set, data is available from `start_time` to now. If this is not set,
  // the `data_availability_interval` in the parent `DataProvider` determines
  // the data availability.
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 122 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfDn-k-9f2k2FniISvU:-NfDn-k-9f2k2FniISvV:b246crp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L122)):*
> ```proto
>   // the `data_availability_interval` in the parent `DataProvider` determines
>   // the data availability.
>   google.type.Interval data_availability_interval = 10;
> ```

nit: Do we want to consider using slightly different names for the two fields, given that they don't have the exact same meaning?

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfDl3A64_5WiLzdHQtK:-NfHRSFhEI6jgbA87q_e:b-kdncbw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*
<details><summary><i>Previously, tristanvuong2021 (Tristan Vuong) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Fair point the requiring end_time. 

Given that only the scheduler cares about these signals we shouldn't see any requisition failure we wouldn't otherwise see.

Arguably at some point we should bake data availability into Report and Metric too.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfCvdKD2MhMH_yoS4aZ:-NfHRoG95xIqZn4XXxJg:b-9ye303) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto#L42)):*

"Update" has specific meaning as a standard method which would have to conform with [AIP-134](https://google.aip.dev/134) and let you update any field. We've instead chosen to use custom methods for setting specific fields.

In terms of "Set" vs. "Replace", we use both verbs for method names in this API. "Replace" was first used for `ReplaceAccountIdentity` due to the fact that the identity had to have already been set. It's also used in the above `ReplaceDataProviderRequiredDuchies` method, which has no such restriction.

It's probably best to be consistent with the "ReplaceDataProviderRequiredDuchies" method since it's on the same resource. The alternative IMO would be to use "Set" and rename the other method, but that's a backwards-incompatible change.
</blockquote></details>

I'm fine keeping Replace on this resource for consistency herein. More broadly we really need to limit the set of verbs we use. Having "Replace" and "Set" methods where the semantics are the same for both verbs is not a good practice. Similarly "Create," "Build," "Make," etc.

This applies at the gRPC interface in nearly critical fashion, but even if our Kotlin/C++ interfaces it'd be nice to be consistent with verb usage.

SanjayVas (changes requested):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: 3 of 4 files reviewed, 5 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 69 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfIaClo9UOp5VmkAbou:-NfIaClp8z9MxcrJ6-fE:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6e455d836e531058cff1ca4a6c0e3b524c9429bc/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L69)):*
> ```proto
> 
>   // Interval for when data is guaranteed to be available for a `Requisition`.
>   // Both `start_time` and `end_time` must be set.
> ```

nit



_Suggestion:_
```proto
If this field is specified, both `start_time` and `end_time` must be set.
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto` line 42 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfCvdKD2MhMH_yoS4aZ:-NfIa_q26TjaMUOdWxij:b-wfpbkv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/data_providers_service.proto#L42)):*
> "Create," "Build," "Make," etc.

FWIW, "create" also has special meaning in gRPC APIs as per [AIP-133](https://google.aip.dev/133). We tend to avoid "make" in Kotlin (though there are a couple exceptions I've been trying to clean up), though it tends to be more common in C++ (e.g. `make_unique`). When not referencing a gRPC method, the "create" verb tends to indicate a factory function. The "build" verb tends to indicate the usage of a builder pattern (e.g. a method on a builder object, or a method which builds using a builder). Note that the builder pattern in Kotlin can be implemented using a DSL.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 125 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfIe0ppE_sjpblajmhu:-NfIe0ppE_sjpblajmhv:brepgxr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6e455d836e531058cff1ca4a6c0e3b524c9429bc/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L125)):*
> ```proto
>   //
>   // If this is not set, the `data_availability_interval` in the parent
>   // `DataProvider` is the interval for when data is guaranteed to be available.
> ```

Do we actually want this behavior? It doesn't seem correct to me. I would assume that if this is not specified then it means there are no guarantees.



_Code quote:_
```proto
  // If this is not set, the `data_availability_interval` in the parent
  // `DataProvider` is the interval for when data is guaranteed to be available.
```

kungfucraig (commented):
> Reviewed 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 125 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfIe0ppE_sjpblajmhu:-NfIf9X_0PHtFpF8VGZl:byygtfe) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6e455d836e531058cff1ca4a6c0e3b524c9429bc/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L125)):*

Do we actually want this behavior? It doesn't seem correct to me. I would assume that if this is not specified then it means there are no guarantees.
</blockquote></details>

We should require that start_time be set.

For end time, it's reasonable to not have it set as we don't want campaigns that are running to have to be updated each day. Of course, we could set a future date, BUT some TV campaigns have no proper end date and can run perpetually.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 118 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfCw-CN0G0DRwcimWqi:-NfE-qGpAaf-B2Wc0jGA:bb77ufg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L118)):*

This field is updated with the standard Update method?
</blockquote></details>

Yes.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 125 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfIe0ppE_sjpblajmhu:-NfIhkS95YNAwv3ahEBl:b-n93rxq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6e455d836e531058cff1ca4a6c0e3b524c9429bc/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L125)):*

We should require that start_time be set.

For end time, it's reasonable to not have it set as we don't want campaigns that are running to have to be updated each day. Of course, we could set a future date, BUT some TV campaigns have no proper end date and can run perpetually.
</blockquote></details>

Changed it, it makes sense that the data should only be guaranteed if it is explicitly stated.

tristanvuong2021 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: 2 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 122 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfDn-k-9f2k2FniISvU:-NfIibJ09Yh2pb5UQjf4:b-7c6npq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L122)):*

nit: Do we want to consider using slightly different names for the two fields, given that they don't have the exact same meaning?
</blockquote></details>

Do you have a possible suggestion?

SanjayVas (approved):
> Reviewed 2 of 2 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 122 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfDn-k-9f2k2FniISvU:-NfIsPOvEpTV5RK-AL1G:b-tu0s31) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/d21ad54e91ecfeb7c3d9c5e4b046e383121bce70/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L122)):*
<details><summary><i>Previously, tristanvuong2021 (Tristan Vuong) wrote…</i></summary><blockquote>

Do you have a possible suggestion?
</blockquote></details>

I think it's actually fine now that we've clarified the meanings.

stevenwarejones (changes requested):
> Reviewed 1 of 2 files at r2, 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWPhVD6JCqZIvyyRpL:-NfWPhVD6JCqZIvyyRpM:b-tud36s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*
> ```proto
>   // Interval for when data is guaranteed to be available for a `Requisition`.
>   // If this field is specified, both `start_time` and `end_time` must be set.
>   google.type.Interval data_availability_interval = 7;
> ```

so, is the idea that a data provider should update this daily (or periodically) as data becomes available? The way our pipeline is set up, we wouldn't fail if the newest data is not availale.... our sketch fulfillment pipeline would just wait on the data. do you think we should support an open-ended data availability for an edp?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 124 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWUg1B6ZY8I2n2eVS3:-NfWUg1B6ZY8I2n2eVS4:bqqbzs2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L124)):*
> ```proto
>   // for when data is guaranteed to be available.
>   //
>   // If this is not set, data is not guaranteed to be available.
> ```

I'm not a fan of this behavior. It should be that if its not set it inherits the parent data_availability_interval

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas, @stevenwarejones, and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWPhVD6JCqZIvyyRpL:-NfWfhoL6gL18dXKnh6G:b-zgylml) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*

so, is the idea that a data provider should update this daily (or periodically) as data becomes available? The way our pipeline is set up, we wouldn't fail if the newest data is not availale.... our sketch fulfillment pipeline would just wait on the data. do you think we should support an open-ended data availability for an edp?
</blockquote></details>

An EDP should be updating this as data becomes available. 

An open-ended availability doesn't make sense because EDP processing lag is non-zero. An event displayed now won't be available until now+lag. Lag of course may be short, but it's non-zero.

The scheduler will want to read this so that it can be managing one in-flight report at a time. Note that the main use case for the scheduler is reporting on live campaigns.

Another use of this is for the Kingdom to monitor EDP SLOs on data processing.

On a side, I think the expected behavior of the EDP if data is not available is undefined.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 124 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWUg1B6ZY8I2n2eVS3:-NfWhNsuDhs3HS9S5gEw:b7yimee) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L124)):*

I'm not a fan of this behavior. It should be that if its not set it inherits the parent data_availability_interval
</blockquote></details>

@stevenwarejones Even if it inherits the parent it's not guaranteed because the availability for a single campaign (e.g. a new campaign) will often start at a time later than the earliest data availability time for an EDP.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 124 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWUg1B6ZY8I2n2eVS3:-NfWhwRY2O9u4qRSW6K8:b-1u128y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L124)):*

@stevenwarejones Even if it inherits the parent it's not guaranteed because the availability for a single campaign (e.g. a new campaign) will often start at a time later than the earliest data availability time for an EDP.
</blockquote></details>

Note that inheritance of the right hand side of the interval is documented below.

kungfucraig (approved):
> Reviewed 2 of 2 files at r4, 1 of 1 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 124 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWUg1B6ZY8I2n2eVS3:-NfXdjezCBag5JBJ8Gre:bhs0k6i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L124)):*

Note that inheritance of the right hand side of the interval is documented below.
</blockquote></details>

I'm still not a super fan of this comment `data is not guaranteed to be available.`. We should have a path to make this field required. And if its not present (because it wasn't required in the past), it inherits the parent (but it should still be guaranteed to be available for the parent)

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 124 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWUg1B6ZY8I2n2eVS3:-NfXvEBWECIQbLWLdKaR:btx075r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L124)):*

I'm still not a super fan of this comment `data is not guaranteed to be available.`. We should have a path to make this field required. And if its not present (because it wasn't required in the past), it inherits the parent (but it should still be guaranteed to be available for the parent)
</blockquote></details>

Yes! We need to make this required after the next release, but we'll need a transition period.

@tristanvuong2021 Maybe we say, "This field will be required in a future release, however until it is required when it is not specified data is not guaranteed to be available."

@stevenwarejones WDYT?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 1 unresolved discussion (waiting on @tristanvuong2021)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWPhVD6JCqZIvyyRpL:-NfXuKHV9GnMl7A1TL-N:b-1i4sy7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*
> our sketch fulfillment pipeline would just wait on the data

> On a side, I think the expected behavior of the EDP if data is not available is undefined.

If would be correct for the EDP to refuse the Requisition with the [UNFULFILLABLE](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e2049532a8da05dc644dd999c5945af2ac5a9e8c/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L197C1-L203C25) Justification, as that is specifically for the case where the EDP *should* be able to fulfill it but cannot. The case where the data *should* be available (which may be governed by an SLA) fits here. It may also be fine to wait, though that behavior may also be governed by an SLA (e.g. that all EDPs should respond to a Requisition by either fulfilling or refusing it within some time frame).

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWPhVD6JCqZIvyyRpL:-Nfl5JOa5HAM7TwVtnBM:bg1m9p1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*

> our sketch fulfillment pipeline would just wait on the data

> On a side, I think the expected behavior of the EDP if data is not available is undefined.

If would be correct for the EDP to refuse the Requisition with the [UNFULFILLABLE](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e2049532a8da05dc644dd999c5945af2ac5a9e8c/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L197C1-L203C25) Justification, as that is specifically for the case where the EDP *should* be able to fulfill it but cannot. The case where the data *should* be available (which may be governed by an SLA) fits here. It may also be fine to wait, though that behavior may also be governed by an SLA (e.g. that all EDPs should respond to a Requisition by either fulfilling or refusing it within some time frame).
</blockquote></details>

the sla use case makes sense. since `google.type.Interval` does not require both fields to be set, would be use a different data type?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 124 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWUg1B6ZY8I2n2eVS3:-Nfl5Dnq42TWEL_zdotZ:b2nign3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L124)):*

Yes! We need to make this required after the next release, but we'll need a transition period.

@tristanvuong2021 Maybe we say, "This field will be required in a future release, however until it is required when it is not specified data is not guaranteed to be available."

@stevenwarejones WDYT?
</blockquote></details>

yep

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 70 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175#-NfWPhVD6JCqZIvyyRpL:-NflUVpl24pXV9rGX1fs:b-nbzjl7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f269f177bb29ae7d79b535ba6f5afc9a981b14f4/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L70)):*
> since `google.type.Interval` does not require both fields to be set, would be use a different data type?

Just documenting that both fields on the submessage must be set for this instance of that type is sufficient. We shouldn't use another type unless it's strictly necessary (e.g. if the type had both fields required and we wanted to make one optional).

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

stevenwarejones (approved):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

stevenwarejones (approved):
> Reviewed 3 of 3 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

stevenwarejones (approved):
> Reviewed 2 of 3 files at r9, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/175)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @tristanvuong2021)

PR #177, opened 2023-09-27
Add signature_algorithm_oid field to SignedData.

SanjayVas:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/crypto.proto` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177#-NfWI2Qp6cZs79Zyt2s8:-NfWI2Qp6cZs79Zyt2s9:b-ijdl3b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e2b2cf0f1b6867658dd446d31fe0b1bc16897f86/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto#L51)):*
> ```proto
>   // If not specified, this is assumed to be the signature algorithm of the
>   // accompanying certificate.
>   string signature_algorithm_oid = 3;
> ```

a little harder to implement but it might be better to only specify the hash algorithm since the encryption algorithm type (rsa vs ec, etc) is already determined by the public key on the certificate.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177)* status: 0 of 1 files reviewed, all discussions resolved

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/crypto.proto` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177#-NfWI2Qp6cZs79Zyt2s8:-NfXEAcy7yRG7qTrqwvQ:b-ft423h) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e2b2cf0f1b6867658dd446d31fe0b1bc16897f86/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto#L51)):*

a little harder to implement but it might be better to only specify the hash algorithm since the encryption algorithm type (rsa vs ec, etc) is already determined by the public key on the certificate.
</blockquote></details>

tl;dr there are actually multiple possible signature algorithms for the same key type and hash algorithm

This is not technically true, though we don't currently support this algorithm as it's a bit more complicated: RSA-PSS. This is a separate signature algorithm for RSA keys that can use the same hash algorithm. You *can* also technically use a key type that's specific to RSA-PSS, but that's not required and it works just fine with a regular RSA key.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`src/main/proto/wfa/measurement/api/v2alpha/crypto.proto` line 51 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177#-NfWI2Qp6cZs79Zyt2s8:-NfXbr_k4mW5fEQHBzt6:b25duq4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e2b2cf0f1b6867658dd446d31fe0b1bc16897f86/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto#L51)):*

tl;dr there are actually multiple possible signature algorithms for the same key type and hash algorithm

This is not technically true, though we don't currently support this algorithm as it's a bit more complicated: RSA-PSS. This is a separate signature algorithm for RSA keys that can use the same hash algorithm. You *can* also technically use a key type that's specific to RSA-PSS, but that's not required and it works just fine with a regular RSA key.
</blockquote></details>

yeah - I agree. lets keep this as-is.

SanjayVas (commented):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/177)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #179, opened 2023-10-04
Add missing patterns for ExchangeStepAttempt.

SanjayVas:
The recurringExchanges resource collection has multiple patterns, so all of its descendents must also support those patterns.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/179)

Issue #181, opened 2023-10-13
API version not indicated for serialized messages


SanjayVas:
The API contains serialized message fields, e.g. for signing and encryption. This is partially due to the fact that protobuf serialization is non-deterministic. Since the messages types are also part of the versioned API, we need an indication of which version of the message was used for serialization. Otherwise a caller using a future API version would not know how to deserialize a message from an older version, e.g. a v2beta caller reading a Measurement that was created in v2alpha.

- [x] #185 
- [x] world-federation-of-advertisers/cross-media-measurement#1306

PR #184, opened 2023-10-13
Indicate API version for serialized fields.

SanjayVas:
See #181.

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto
@@ -54,6 +54,12 @@ message DataProvider {
   // verified using `certificate`.
   SignedData public_key = 4 [(google.api.field_behavior) = REQUIRED];
 
+  // Version of the API that `public_key` was last updated in.
+  //
+  // This determines which version of a message definition is used for
+  // serialized messages.
+  string public_key_api_version = 7 [(google.api.field_behavior) = OUTPUT_ONLY];
+
   // Display name of the data provider.
   string display_name = 5;
 
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
@@ -41,6 +41,12 @@ message EventGroup {
   // Canonical format: dataProviders/{data_provider}/eventGroups/{event_group}
   string name = 1;
 
+  // Version of the API that this resource was last updated in.
+  //
+  // This determines which version of a message definition is used for
+  // serialized messages.
+  string api_version = 10 [(google.api.field_behavior) = OUTPUT_ONLY];
+
   // Resource name of the `MeasurementConsumer` associated with this
   // `EventGroup`.
   string measurement_consumer = 2 [
--- a/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/exchange_step.proto
@@ -68,6 +68,12 @@ message ExchangeStep {
     ];
   }
 
+  // Denormalized `api_version` field from the ancestor `RecurringExchange`.
+  string api_version = 8 [
+    (google.api.field_behavior) = OUTPUT_ONLY,
+    (google.api.field_behavior) = IMMUTABLE
+  ];
+
   // Serialized denormalized `exchange_workflow` field from the ancestor
   // `RecurringExchange`.
   bytes serialized_exchange_workflow = 3 [
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
@@ -20,8 +20,8 @@ import "google/api/field_behavior.proto";
 import "google/api/resource.proto";
 import "google/protobuf/duration.proto";
 import "wfa/measurement/api/v2alpha/crypto.proto";
-import "wfa/measurement/api/v2alpha/protocol_config.proto";
 import "wfa/measurement/api/v2alpha/direct_computation.proto";
+import "wfa/measurement/api/v2alpha/protocol_config.proto";
 
 option java_package = "org.wfanet.measurement.api.v2alpha";
 option java_multiple_files = true;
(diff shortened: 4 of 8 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184)

stevenwarejones (changes requested):
Doesn't this place a burden on the party deserializing data to need to know multiple API versions' messages? Instead, why not just always store serialized data in a pair with its FileDescriptorSet?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-:-Ngfv7v_-0sG8-N4Xjt_:b-rgxx50)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 285 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-NgfuiX3-9s3pfCIC_Ie:-NgfuiX3-9s3pfCIC_If:b-dvqarx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/4d43e03606763226a709b86f3359367e1750e295/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L285)):*
> ```proto
> 
>   // Output message for a `Result`.
>   message ResultOutput {
> ```

not a huge fan of this name. Maybe just EncryptedResult?

SanjayVas (commented):
Yes, though this should ideally only be a temporary state during version transitions. The API requires that when you serialize messages you use the message version from the API you're calling. Note that this design decision was already made: we already have most of these versions in the internal API, which we persist in the DB already. We just didn't end up exposing them in the public API, only in the internal and system APIs.

> Instead, why not just always store serialized data in a pair with its FileDescriptorSet?

Use of FileDescriptorSet (self-describing messages) is for cases when there isn't a protobuf definition in a well-known location. It's worse for multiple reasons:

1. It duplicates the protobuf descriptor in every response
2. Client code does not have access to the generated language-specific code

We only do this for EventGroup metadata, as that is EDP-specific. Contrast this with Event templates, which are shared across an entire market (Kingdom instance). For the latter, we expect all clients to have access to the event template protobuf definitions.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-:-Ngt0aDb0njozznIaHMe:b-q8jva6)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 285 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-NgfuiX3-9s3pfCIC_Ie:-Ngt31Wz2Uj5jCeajGiV:b9vdhu1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/4d43e03606763226a709b86f3359367e1750e295/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L285)):*

not a huge fan of this name. Maybe just EncryptedResult?
</blockquote></details>

I thought of that name, but the encrypted result is one of the values that this message contains. This is really trying to capture everything the API needs for the output-only field.

I'm of course open to other naming suggestions.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/recurring_exchange.proto` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-NgtAW0JE3YajclOHKm1:-NgtAW0JE3YajclOHKm2:b-du8jyx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/4d43e03606763226a709b86f3359367e1750e295/src/main/proto/wfa/measurement/api/v2alpha/recurring_exchange.proto#L48)):*
> ```proto
>   // This determines which version of a message definition is used for
>   // serialized messages.
>   string api_version = 7 [
> ```

i'd prefer these changes be in the api_version pr

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184)* status: 0 of 8 files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/recurring_exchange.proto` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-NgtAW0JE3YajclOHKm1:-NgtImwC0Niy5pNB6iCu:b-8ayi6t) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/4d43e03606763226a709b86f3359367e1750e295/src/main/proto/wfa/measurement/api/v2alpha/recurring_exchange.proto#L48)):*

i'd prefer these changes be in the api_version pr
</blockquote></details>

Can you clarify what you mean? This is the PR for adding API version fields.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184)* status: 0 of 8 files reviewed, 1 unresolved discussion

___
*[`src/main/proto/wfa/measurement/api/v2alpha/recurring_exchange.proto` line 48 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/184#-NgtAW0JE3YajclOHKm1:-Ngu7Vsq9ba00CTlN52s:b-liovw6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/4d43e03606763226a709b86f3359367e1750e295/src/main/proto/wfa/measurement/api/v2alpha/recurring_exchange.proto#L48)):*

Can you clarify what you mean? This is the PR for adding API version fields.
</blockquote></details>

maybe a reviewable bug? or maybe I was just getting the tabs confused.

PR #178, opened 2023-09-28
Refactor custom direct methodology

riemanli:
## TL;DR
This pull request introduces changes to the `direct_computation.proto` and `protocol_config.proto` files. The changes include the addition of a new message `FrequencyVariances` and a `oneof` field `variance` in the `CustomDirectMethodology` message. It also modifies the comment for `custom_direct_methodology` in `ProtocolConfig`.
> 
## What changed
In `direct_computation.proto`, the `variance` field in `CustomDirectMethodology` has been replaced with a `oneof` field that can either be a `scalar` or `FrequencyVariances`. The `FrequencyVariances` message includes two maps for `variances` and `k_plus_variances`, which represent the variances of different types of frequency distributions from frequency 1 to the maximum frequency specified in the measurement spec.

> 
In `protocol_config.proto`, the comment for `custom_direct_methodology` has been updated to include a requirement for data providers to guarantee the independence of the result by salting their hash functions.
> 
## Why make this change
This change allows for more flexibility in specifying the variance in the `CustomDirectMethodology` message. It also clarifies the requirements for data providers who choose to use a custom direct methodology.

PR #186, opened 2023-10-23
Make CustomDirectMethodology.variance not required

riemanli:
TV measurements will not have variances, so the Kingdom API needs to allow that by either
1. allowing the data provider not indicating the variance of their direct measurement, i.e. `variance` is not required,
2. use `NaN` value in `scalar` or `frequency`, or
3. adding a new type `Unknown` under `variance`.

This PR proposed to use option 1.

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
@@ -24,31 +24,58 @@ option java_outer_classname = "DirectComputationProto";
 
 // Information about the custom direct methodology.
 message CustomDirectMethodology {
-  // Different types of variances of a frequency distribution result.
-  message FrequencyVariances {
-    // The variances of a frequency distribution from frequency 1 to maximum
-    // frequency specified in the measurement spec.
-    map<int64, double> variances = 1 [(google.api.field_behavior) = REQUIRED];
-    // The variances of a k+ frequency distribution from frequency 1 to
-    // maximum frequency specified in the measurement spec.
-    //
-    // A K+ frequency distribution is derived from the frequency distribution by
-    // calculating the reach ratio of frequency K and above, i.e. reversed
-    // cumulative sum of the frequency distribution. For example, a frequency
-    // distribution {1: 0.4, 2: 0.3, 3: 0.2, 4:0.1, 5: 0.0} will have a K+
-    // frequency distribution {1: 1.0, 2: 0.6, 3: 0.3, 4:0.1, 5: 0.0}.
-    map<int64, double> k_plus_variances = 2
-        [(google.api.field_behavior) = REQUIRED];
+  // The information about a variance.
+  message Variance {
+    // Different types of variances of a frequency distribution result.
+    message FrequencyVariances {
+      // The variances of a frequency distribution from frequency 1 to maximum
+      // frequency specified in the measurement spec.
+      map<int64, double> variances = 1 [(google.api.field_behavior) = REQUIRED];
+      // The variances of a k+ frequency distribution from frequency 1 to
+      // maximum frequency specified in the measurement spec.
+      //
+      // A K+ frequency distribution is derived from the frequency distribution
+      // by calculating the reach ratio of frequency K and above, i.e. reversed
+      // cumulative sum of the frequency distribution. For example, a frequency
+      // distribution {1: 0.4, 2: 0.3, 3: 0.2, 4:0.1, 5: 0.0} will have a K+
+      // frequency distribution {1: 1.0, 2: 0.6, 3: 0.3, 4:0.1, 5: 0.0}.
+      map<int64, double> k_plus_variances = 2
+          [(google.api.field_behavior) = REQUIRED];
+    }
+
+    // Information about lack of variance.
+    message Unavailable {
+      // Reason for a `Unavailable`.
+      enum Reason {
+        // Default value used if the reason is omitted.
+        //
+        // Used to capture unset reason which is invalid. This enum constant
+        // should never be set.
+        REASON_UNSPECIFIED = 0;
+        // When the variance is mathematically not derivable from a custom
+        // direct methodology.
+        UNDERIVABLE = 1;
+        // When the variance is obtained by upstream providers and not
+        // accessible.
+        INACCESSIBLE = 2;
+      }
+      // Reason for this `Unavailable`.
+      Reason reason = 1 [(google.api.field_behavior) = REQUIRED];
+    }
+
+    // The type of variance associated with a result. Required.
+    oneof type {
+      // The variance when the computation result is a scalar type.
+      double scalar = 1;
+      // The variance when the computation result is a frequency type.
+      FrequencyVariances frequency = 2;
+      // The variance is unavailable for a custom direct methodology.
+      Unavailable unavailable = 3;
+    }
   }
 
-  // The variance of the result computed from the custom direct methodology.
-  // REQUIRED.
-  oneof variance {
-    // The variance when the result is a scalar type.
-    double scalar = 1;
-    // The variance when the result is a frequency type.
-    FrequencyVariances frequency = 2;
-  }
+  // The variance of the result computed from this custom direct methodology.
+  Variance variance = 1 [(google.api.field_behavior) = REQUIRED];
 }
 
 // Parameters used when applying the deterministic count distinct methodology.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186)

riemanli:
Current dependencies on/for this PR:
* main
  * **PR #186** <a href="https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement-api/186" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/world-federation-of-advertisers/cross-media-measurement-api/186?utm_source=stack-comment).

SanjayVas (commented):
I thought we explicitly didn't want to use option 1 and instead decided on option 2. @kungfucraig 

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NhU51ak4LPpwWrs48F1:b-h54sol)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @kungfucraig)

SanjayVas (commented):
To further elaborate, I thought we were going to make the following changes to the public API:

1. Document the use of NaN to indicate cases where variance is explicitly unknown/unknowable. We prefer this over simply allowing a variance to not be specified due to the fact that it's explicit and doesn't just e.g. reflect an EDP forgetting to specify the known variance.
2. Document that `0` is a valid value for `scalar` variance just to make it extra clear since proto3 does not normally differentiate between "unset" and "set to default value". This is to allow the theoretical case where we want to treat a result as having a known 0 variance, as opposed to the case where the variance is unknown.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NhU5Osl7Q7w8r3TfMNO:b-heh4fr)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @kungfucraig)

kungfucraig (commented):
I talked to Rieman about this, and he wanted to see what it looked like if we went this way.

However, if we go this way, we need to box the scalar, so that we can distinguish between not set and zero.

So the trade-off is do we box the primitive value or make NaN the "unknown variance" value. 

When I talked with Sanjay, we thought the NaN route was better, especially as it related to development cost and backward compatibility. And it seemed no worse than boxing the value wrt usability.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni5gX6i9dDbxbtD8E4u:b-r7b8ma)* status: 0 of 1 files reviewed, all discussions resolved

SanjayVas (commented):
I originally thought we had to use the optional keyword for the scalar, but I don't think we actually do as it's in a oneof. The oneof case will simply be "not set".

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni5sUFk2PBjnYqqyOra:b-1jbaey)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @kungfucraig)

kungfucraig (commented):
So maybe this is the change we want? It takes care of frequency at the same time.

If we do want this, a bit more documentation about what it means for this to be unset would be nice.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni5udkM0jLE6Sf4IvBZ:by7a0ht)* status: 0 of 1 files reviewed, all discussions resolved

SanjayVas (commented):
For background on the scalar issue, it's that we need a way to support a known variance of `0` while also still enforcing requirements. For a regular scalar field, there is no difference between `0` and unset, meaning setting a required scalar field to `0` is an error. In this case though, the scalar is wrapped in a oneof and it's the oneof that's required. I believe that proto3 will treat an explicit `0` as the oneof being set.

With that said, here are the options:

1. Require methodology, but allow NaN for custom methodology to explicitly indicate that the variance is invalid (i.e. unknowable).
2. Don't require methodology at all. If custom methodology is specified, it must include a valid variance.
3. Require methodology, but don't require a variance to be specified for custom methodology.

(1) requires that the EDP be explicit about the variance being a custom methodology with an unknowable variance. It does this at the cost of using `NaN` as a sentinel value. (3) has the EDP be explicit that the methodology is custom, but not require them to indicate a variance. (2) is basically the status quo.

There are tradeoffs for all of these. I believe this PR does (3). I'm okay with any of the options as long as we understand the implications.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni6Bfcz9SYGk-d4LwTE:b217amt)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @kungfucraig)

kungfucraig (commented):
Agree this PR does option 3, and it's fine.

Option 1 is fine.

Option 2 is not distinguishable from the EDP failing to implement the contract, which I don't like.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni6UvZE1inoHPoo90l5:bx4sdlx)* status: 0 of 1 files reviewed, all discussions resolved

kungfucraig (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

stevenwarejones (changes requested):
Aren't options 2 and options 3 both possibly indicative of the EDP failing to implement the contract?

There's also the use case where its a histogram result but the EDP doesn't know or can't know the variances from their methodology. In that case, they should also not populate the oneof frequency value?  Or Option 1 would require a histogram type sentinel value as well within frequency in this case. I don't like that option.

I can't say I'm a huge fan of any of these options. 

What about if we wrap the oneof inside a VarianceType Message? I think the contract becomes clearer if the EDP includes the VarianceType message but then doesn't populate fields within that they don't know the variance.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NiGuJfz2Zb_QZYMZjsD:b-8m7g15)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiGurmY2m_gF-DgbUDc:-NiGurmY2m_gF-DgbUDd:bukeaa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cd539f4a6684ff70bd4eab41573c01db3fae82b8/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L48)):*
> ```proto
>   // If the oneof is not set, it indicates the variance is unknown or not
>   // derivable from the custom methodology.
>   oneof variance {
> ```

One problem i have with the current naming here is that we aren't including all variances.... just the variance associated with constructing a scalar result from a list of VIDs. Maybe update the comment around that?

kungfucraig (commented):
What I meant by failing to implement the contract was with respect to moving from the previous version of the API to the new version. With option 3, the EDP at least has to construct the Custom Methodology message, whereas option 2 would result if the EDP took no action (i.e. didn't acknowledge the API update).

To some extent we're also trying to minimize dev work, and option 3 should be a documentation only type of change on the Kingdom side.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NiH4QZ424-_vRuEO_RC:b-glphda)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @riemanli and @SanjayVas)

riemanli:
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiGurmY2m_gF-DgbUDc:-NiHBxASDTOipLk4Jgfd:b-d86xi4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cd539f4a6684ff70bd4eab41573c01db3fae82b8/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L48)):*
> One problem i have with the current naming here is that we aren't including all variances.... just the variance associated with constructing a scalar result from a list of VIDs.

I am not sure what you meant. Could you elaborate? The oneof has two types of variances. One is scalar, and the other is Frequency. The scalar type can be used for Reach, Impression, and Duration results.

riemanli (commented):
I modified based on the feedback:
1. The field `variance` in `CustomDirectMethodology` is now a `Variance` type.
2. `FrequencyVariances` and `Unavailable` are defined inside `Variance`.
3. `Unavailable` has a field `reason` to indicate why a variance is not available. Currently there are two reasons in the enum --`NOT_DERIVABLE` and `NO_ACCESS`. Can add more if there are more use cases.
4. `Variance` has an `oneof type` that indicate which type of variance it is.

In this design, we won't break the protobuf as the field name doesn't change, and we are able to indicate the unavailability  of variance with no grey area. In terms of dev effort, there are no change in the kingdom and only a few changes in the reporting server.

Will modify the PR description once it's approved to avoid redundant editing.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NiH6FKr8jh0r36cEAAN:bhseu)* status: 0 of 1 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

stevenwarejones (changes requested):
i like this change.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NiK0nwJ7XsYI9ORsC6R:b-m320xj)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiGurmY2m_gF-DgbUDc:-NiK0R-n4JwIrHxc7Z8M:b-nlayiz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cd539f4a6684ff70bd4eab41573c01db3fae82b8/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L48)):*
<details><summary><i>Previously, riemanli (Rieman) wrote…</i></summary><blockquote>

> One problem i have with the current naming here is that we aren't including all variances.... just the variance associated with constructing a scalar result from a list of VIDs.

I am not sure what you meant. Could you elaborate? The oneof has two types of variances. One is scalar, and the other is Frequency. The scalar type can be used for Reach, Impression, and Duration results.
</blockquote></details>

there's additional variance for example, from VID model construction. We should clarify which sources of variance should be included.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 50 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiK0zcICv1yH7FKtDcJ:-NiK0zcICv1yH7FKtDcK:b-r9wu6k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L50)):*
> ```proto
>       // Reason for a `Unavailable`.
>       enum Reason {
>         // Default value used if the reason is omitted.
> ```

Are we going to allow UNSPECIFIED as a valid choice? I thought in general we were avoiding allowing UNSPECIFIED to ever be a valid choice.

kungfucraig (changes requested):
I think we want to maintain backward compatibility here, which I think can be done if we drop the Variance message and keep the oneof as it was before.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NiKgjIA76ChdCq_8LQa:b-tmme6i)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @riemanli and @SanjayVas)

stevenwarejones:
I’d prefer a breaking change. Who is using this now?

riemanli (commented):
Note that we already made changes to `variance` once in #178 on Oct 4th after the major release v0.4.0.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-NiL4HJACi7kPKF5DuiR:b-481c39)* status: 0 of 1 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiGurmY2m_gF-DgbUDc:-NiLWcwR2UqzbJp0jvpR:b42o71k) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cd539f4a6684ff70bd4eab41573c01db3fae82b8/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L48)):*

there's additional variance for example, from VID model construction. We should clarify which sources of variance should be included.
</blockquote></details>

The methodology here is only used in measurement computation. This is the variance of a measurement result. They should consider all sources that can impact the result to compute the variance.
If it helps, I can add a comment stating that "all sources that impact the result should be considered in the variance calculation."
___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 50 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiK0zcICv1yH7FKtDcJ:-NiL80yxAr5uchdo1Tlt:bx7z06) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L50)):*

Are we going to allow UNSPECIFIED as a valid choice? I thought in general we were avoiding allowing UNSPECIFIED to ever be a valid choice.
</blockquote></details>

We will raise an error if UNSPECIFIED is set. Without it, we don't know if a valid option is set by the user or by default as enum has no explicit presence.

SanjayVas (approved):
> Reviewed 1 of 1 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @riemanli, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 50 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiK0zcICv1yH7FKtDcJ:-NiMUh2N5t4GEQM72JYs:b-ukyfuv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L50)):*
<details><summary><i>Previously, riemanli (Rieman) wrote…</i></summary><blockquote>

We will raise an error if UNSPECIFIED is set. Without it, we don't know if a valid option is set by the user or by default as enum has no explicit presence.
</blockquote></details>

To elaborate on Rieman's answer, if a scalar field is REQUIRED then it is an error for it to be set to its default value. The default value for an enum field is `0`. Therefore API enums tend have an UNSPECIFIED entry for the default value (`0`). 
___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiMVjkL2e8_xxob4UXW:-NiMVjkL2e8_xxob4UXX:be98pcj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L57)):*
> ```proto
>         // When the variance is obtained by upstream providers and not
>         // accessible.
>         NO_ACCESS = 2;
> ```

nit: Values have different forms. UNDERIVABLE and INACCESSIBLE would make them match



_Suggestion:_
```proto
        // When the variance is mathematically not derivable from a custom
        // direct methodology.
        UNDERIVABLE = 1;
        // When the variance is obtained by upstream providers and not
        // accessible.
        INACCESSIBLE = 2;
```

riemanli (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 57 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiMVjkL2e8_xxob4UXW:-NiMWwXxAhBd6VsDKBCC:bt4ox15) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L57)):*

nit: Values have different forms. UNDERIVABLE and INACCESSIBLE would make them match
</blockquote></details>

Done. Thanks for the suggestion!

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 48 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiGurmY2m_gF-DgbUDc:-Ni_-NbQERTEMI9SjYg_:b-b81g73) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cd539f4a6684ff70bd4eab41573c01db3fae82b8/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L48)):*
<details><summary><i>Previously, riemanli (Rieman) wrote…</i></summary><blockquote>

The methodology here is only used in measurement computation. This is the variance of a measurement result. They should consider all sources that can impact the result to compute the variance.
If it helps, I can add a comment stating that "all sources that impact the result should be considered in the variance calculation."
</blockquote></details>

I don't want to overthink this. I think we'll need to revisit this in the future.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 50 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiK0zcICv1yH7FKtDcJ:-Ni_-3ksEDpK6cCxUAwU:bn8uwog) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L50)):*

To elaborate on Rieman's answer, if a scalar field is REQUIRED then it is an error for it to be set to its default value. The default value for an enum field is `0`. Therefore API enums tend have an UNSPECIFIED entry for the default value (`0`).
</blockquote></details>

The comment doesn't suggest its an invalid choice. We should add a comment that reason should never be set to`REASON_UNSPECIFIED`.

kungfucraig (approved):
Okay, I think it means we don't want anyone to migrate to this before 0.5 then, right?

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni_KY8xEGQ-0pBJebkU:bnsvafe)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @riemanli)

riemanli (commented):
Yes

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-:-Ni_aitH702yEg5T0HhV:b2nhswy)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 50 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186#-NiK0zcICv1yH7FKtDcJ:-Ni_afeTDt_rIcCSILTC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/5b6babf2052d561340239f13dd3c9925ebc7f5f2/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L50)):*

The comment doesn't suggest its an invalid choice. We should add a comment that reason should never be set to`REASON_UNSPECIFIED`.
</blockquote></details>

Done.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/186)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @riemanli)

PR #197, opened 2024-02-01
Add custom_maximum_frequency_per_user field into DeterministicCount methodology

iverson52000:
- Add `custom_maximum_frequency_per_user` field for EDP to report back the maximumFrequencyCap they use for impression measurement
- Add documentation that if EDP specify the new field for impression, the `maximum_frequency_per_user` in `measurement_spec` will be ignored

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
@@ -85,12 +85,10 @@ message DeterministicCountDistinct {}
 message DeterministicDistribution {}
 
 // Parameters used when applying the deterministic count methodology.
-message DeterministicCount {}
-
-// Parameters used when applying the Dynamic Clipping count methodology.
-message DynamicClipCount {
-  // Optimized maximum frequency per user value calculated by the DataProvider.
-  int32 maximum_frequency_per_user = 1 [(google.api.field_behavior) = REQUIRED];
+message DeterministicCount {
+  // Optimized maximum frequency per user value calculated by the
+  // DataProvider.
+  int32 dynamic_clipping_maximum_frequency_per_user = 1;
 }
 
 // Parameters used when applying the deterministic sum methodology.
--- a/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
@@ -86,9 +86,10 @@ message DeterministicDistribution {}
 
 // Parameters used when applying the deterministic count methodology.
 message DeterministicCount {
-  // Optimized maximum frequency per user value calculated by the
-  // DataProvider.
-  int32 dynamic_clipping_maximum_frequency_per_user = 1;
+  // Custom maximum frequency per user value calculated by the
+  // DataProvider. If this is specified, the maximum_frequency_per_user in
+  // measurement_spec will be ignored.
+  int32 custom_maximum_frequency_per_user = 1;
 }
 
 // Parameters used when applying the deterministic sum methodology.
--- a/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
@@ -87,6 +87,12 @@ message DeterministicDistribution {}
 // Parameters used when applying the deterministic count methodology.
 message DeterministicCount {}
 
+// Parameters used when applying the Dynamic Clipping count methodology.
+message DynamicClipCount {
+  // Optimized maximum frequency per user value calculated by the EDP.
+  int32 maximum_frequency_per_user = 1;
+}
+
 // Parameters used when applying the deterministic sum methodology.
 message DeterministicSum {}
 
--- a/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto
@@ -89,8 +89,8 @@ message DeterministicCount {}
 
 // Parameters used when applying the Dynamic Clipping count methodology.
 message DynamicClipCount {
-  // Optimized maximum frequency per user value calculated by the EDP.
-  int32 maximum_frequency_per_user = 1;
+  // Optimized maximum frequency per user value calculated by the DataProvider.
+  int32 maximum_frequency_per_user = 1 [(google.api.field_behavior) = REQUIRED];
 }
 
 // Parameters used when applying the deterministic sum methodology.
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
@@ -234,6 +234,9 @@ message Measurement {
 
         // Deterministic count methodology.
         DeterministicCount deterministic_count = 4;
+
+        // Dynamic Clipping count methodology.
+        DynamicClipCount dynamic_clip_count = 5;
       }
     }
     // The impression result.
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
@@ -234,9 +234,6 @@ message Measurement {
 
         // Deterministic count methodology.
         DeterministicCount deterministic_count = 4;
-
-        // Dynamic Clipping count methodology.
-        DynamicClipCount dynamic_clip_count = 5;
       }
     }
     // The impression result.
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -94,9 +94,9 @@ message MeasurementSpec {
 
     // Maximum frequency per user that will be included in this measurement.
     //
-    // If the DynamicClipCount methodology is used,
-    // this field will be ignored and the value from that message will be used
-    // instead.
+    // If the dynamic_clip_cap in the DeterministicCount methodology is
+    // specified, this field will be ignored and the value from that message
+    // will be used instead.
     int32 maximum_frequency_per_user = 2
         [(google.api.field_behavior) = REQUIRED];
   }
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -92,11 +92,11 @@ message MeasurementSpec {
     // Differential privacy parameters.
     DifferentialPrivacyParams privacy_params = 1;
 
-    // Maximum frequency per user that will be included in this measurement.
+    // Maximum frequency per user that would be included in this measurement.
     //
-    // If the dynamic_clip_cap in the DeterministicCount methodology is
-    // specified, this field will be ignored and the value from that message
-    // will be used instead.
+    // If the custom_maximum_frequency_per_user in the DeterministicCount
+    // methodology is specified, this field will be ignored and the value from
+    // that message will be used instead.
     int32 maximum_frequency_per_user = 2
         [(google.api.field_behavior) = REQUIRED];
   }
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -93,6 +93,10 @@ message MeasurementSpec {
     DifferentialPrivacyParams privacy_params = 1;
 
     // Maximum frequency per user that will be included in this measurement.
+    //
+    // Required and validated at Kingdom. It could be ignored by EDP if EDP uses
+    // Dynamic Clipping methodology to calculate an optimized value and report
+    // back to Kingdom.
     int32 maximum_frequency_per_user = 2;
   }
 
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto
@@ -94,10 +94,11 @@ message MeasurementSpec {
 
     // Maximum frequency per user that will be included in this measurement.
     //
-    // Required and validated at Kingdom. It could be ignored by EDP if EDP uses
-    // Dynamic Clipping methodology to calculate an optimized value and report
-    // back to Kingdom.
-    int32 maximum_frequency_per_user = 2;
+    // If the DynamicClipCount methodology is used,
+    // this field will be ignored and the value from that message will be used
+    // instead.
+    int32 maximum_frequency_per_user = 2
+        [(google.api.field_behavior) = REQUIRED];
   }
 
   // Parameters for a duration measurement.
--- a/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
@@ -118,6 +118,8 @@ message ProtocolConfig {
     message DeterministicDistribution {}
     // Configuration parameters for the deterministic count methodology.
     message DeterministicCount {}
+    // Configuration parameters for the Dynamic Clipping count methodology.
+    message DynamicClipCount {}
     // Configuration parameters for the deterministic sum methodology.
     message DeterministicSum {}
     // Configuration parameters for the direct Liquid Legions distribution
@@ -176,6 +178,11 @@ message ProtocolConfig {
     //
     // Can be used in frequency computations.
     LiquidLegionsDistribution liquid_legions_distribution = 8;
+
+    // Dynamic Clipping count methodology.
+    //
+    // Can be used in impression computations.
+    DynamicClipCount dynamic_clip_count = 9;
   }
 
   // Configuration for the Reach-Only Liquid Legions v2 protocol.
--- a/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/protocol_config.proto
@@ -118,8 +118,6 @@ message ProtocolConfig {
     message DeterministicDistribution {}
     // Configuration parameters for the deterministic count methodology.
     message DeterministicCount {}
-    // Configuration parameters for the Dynamic Clipping count methodology.
-    message DynamicClipCount {}
     // Configuration parameters for the deterministic sum methodology.
     message DeterministicSum {}
     // Configuration parameters for the direct Liquid Legions distribution
@@ -178,11 +176,6 @@ message ProtocolConfig {
     //
     // Can be used in frequency computations.
     LiquidLegionsDistribution liquid_legions_distribution = 8;
-
-    // Dynamic Clipping count methodology.
-    //
-    // Can be used in impression computations.
-    DynamicClipCount dynamic_clip_count = 9;
   }
 
   // Configuration for the Reach-Only Liquid Legions v2 protocol.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 5 unresolved discussions (waiting on @iverson52000, @kungfucraig, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 92 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaPcrb-pg978ezT3MJ:-NpaPcrb-pg978ezT3MK:bf6hkyx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L92)):*
> ```proto
> // Parameters used when applying the Dynamic Clipping count methodology.
> message DynamicClipCount {
>   // Optimized maximum frequency per user value calculated by the EDP.
> ```

We don't use the term EDP in the API. Instead we refer to the `DataProvider` resource



_Suggestion:_
```proto
`DataProvider`
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaPpgl4YzFAEbebIsJ:-NpaPpgl4YzFAEbebIsK:b-ku28t2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L93)):*
> ```proto
> message DynamicClipCount {
>   // Optimized maximum frequency per user value calculated by the EDP.
>   int32 maximum_frequency_per_user = 1;
> ```

Indicate that this is required



_Suggestion:_
```proto
int32 maximum_frequency_per_user = 1 [(google.api.field_behavior) = REQUIRED];
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaQ2b0E5Tw-SKhsG4P:-NpaQ2b0E5Tw-SKhsG4Q:bfkqnb1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*
> ```proto
>     // Maximum frequency per user that will be included in this measurement.
>     //
>     // Required and validated at Kingdom. It could be ignored by EDP if EDP uses
> ```

Drop this comment in favor of a field behavior annotation (`[(google.api.field_behavior) = REQUIRED]`)



_Code quote:_
```proto
Required and validated at Kingdom. 
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaQ9_C5wuLttpxQJCA:-NpaQ9_C5wuLttpxQJCB:b-jdvdq7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*
> ```proto
>     // Maximum frequency per user that will be included in this measurement.
>     //
>     // Required and validated at Kingdom. It could be ignored by EDP if EDP uses
> ```

Same note here re: EDP



_Code quote:_
```proto
EDP
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 99 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaQGeHDhVIUHmd_6Cm:-NpaQGeHDhVIUHmd_6Cn:b-f0jvpo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L99)):*
> ```proto
>     // Required and validated at Kingdom. It could be ignored by EDP if EDP uses
>     // Dynamic Clipping methodology to calculate an optimized value and report
>     // back to Kingdom.
> ```

Don't reference the Kingdom here and instead specifically reference the relevant parts of the API

e.g. "If the `DynamicClipCount` methodology is used, this field will be ignored and the value from that message will be used instead."



_Code quote:_
```proto
    // Dynamic Clipping methodology to calculate an optimized value and report
    // back to Kingdom.
```

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 2 of 4 files reviewed, 5 unresolved discussions (waiting on @kungfucraig, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 92 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaPcrb-pg978ezT3MJ:-Npeq_qZ0g3Z1vWHNuQq:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L92)):*

We don't use the term EDP in the API. Instead we refer to the `DataProvider` resource
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 93 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaPpgl4YzFAEbebIsJ:-Npeqat73dtmB-iKkpTB:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L93)):*

Indicate that this is required
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaQ9_C5wuLttpxQJCA:-Npeqe5V-seBGcPnfENs:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*

Same note here re: EDP
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaQ2b0E5Tw-SKhsG4P:-NpeqeIv9JX9omgBjJmC:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*

Drop this comment in favor of a field behavior annotation (`[(google.api.field_behavior) = REQUIRED]`)
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 99 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NpaQGeHDhVIUHmd_6Cm:-Npeqja826bXfgzYxUpo:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/65181ddedf0051d31684551ed71209bfeaf6e678/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L99)):*

Don't reference the Kingdom here and instead specifically reference the relevant parts of the API

e.g. "If the `DynamicClipCount` methodology is used, this field will be ignored and the value from that message will be used instead."
</blockquote></details>

Done.

SanjayVas (approved):
> Reviewed 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @riemanli)

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 101 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Npf_PpQ0UmPSTPImDxx:-Npf_PpQ0UmPSTPImDxy:b-8dffcn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L101)):*
> ```proto
>     // instead.
>     int32 maximum_frequency_per_user = 2
>         [(google.api.field_behavior) = REQUIRED];
> ```

FYI for future reviewers: this field has always been required. This PR is just fixing the documentation of that fact.



_Code quote:_
```proto
(google.api.field_behavior) = REQUIRED
```

riemanli (approved):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 101 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Npf_PpQ0UmPSTPImDxx:-NpuLd1-0wDOs0_JkGWB:b-uv4tn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L101)):*

FYI for future reviewers: this field has always been required. This PR is just fixing the documentation of that fact.
</blockquote></details>

Thanks for the clarification!

kungfucraig (approved):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r2, 2 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @Marco-Premier and @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 1 of 4 files at r1, 1 of 2 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @iverson52000 and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-NputLm64JQ0-yGj1pLe:bsrr06u) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
> ```proto
>         DeterministicCount deterministic_count = 4;
> 
>         // Dynamic Clipping count methodology.
> ```

Is dynamic clip count a methodology? Can it be done in a non-deterministic manner with a custom variance? If so, then it should be a new field rather than inside of methodology.

Also, it doesn't have to be dynamic, either. For example, an EDP might have a requirement of a different constant clipping value than what the MC provides.

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-Npuv_IgCnUnbMp963O6:b9lqlra) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
> an EDP might have a requirement of a different constant clipping value than what the MC provides.

I'm not sure we care about or want to support this use case. @kungfucraig ?

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-Npv1EEG2LMWbercGvkP:b-7l0tel) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*

> an EDP might have a requirement of a different constant clipping value than what the MC provides.

I'm not sure we care about or want to support this use case. @kungfucraig ?
</blockquote></details>

Hi Steven. If I understand your question correctly, the new `DynamicClipCount` methodology is EDP uses the `DynamicCliping` [library](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/eventdataprovider/differentialprivacy/DynamicClipping.kt) to compute the noised impression value and optimized threshold(`maximum_frequency_per_user`) and report back to Kingdom. In this case, it’s considered as a methodology since it’s different from other ones, ex. `DeterministicCount` which uses fixed cap passed from MC?

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @Marco-Premier and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-NpvBTcF8MoGNpj_R5vv:bcqege1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

Hi Steven. If I understand your question correctly, the new `DynamicClipCount` methodology is EDP uses the `DynamicCliping` [library](https://github.com/world-federation-of-advertisers/cross-media-measurement/blob/main/src/main/kotlin/org/wfanet/measurement/eventdataprovider/differentialprivacy/DynamicClipping.kt) to compute the noised impression value and optimized threshold(`maximum_frequency_per_user`) and report back to Kingdom. In this case, it’s considered as a methodology since it’s different from other ones, ex. `DeterministicCount` which uses fixed cap passed from MC?
</blockquote></details>

@stevenwarejones I'm not exactly sure what you're proposing, can you explain a bit?

My sense is that keeping DC on its own allows the system to infer some things about it automatically, which allows EDPs to offload those calculations to Halo. 

If we capture it more generally it'll mean more work for EDPs I think.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw, @jiayu-google, and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-NpzZ5kS22u9VhyMqeXf:bif2im0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*

@kungfucraig - the dynamic clipping value has to be calculated by the EDP, per my understanding.
@iverson52000 , I see the following scenarios:
1. EDP calculates dynamic clipping to maximize utility vs privacy budget
2. EDP uses a different universal max than what the kingdom provides, so it uses that value
Once Custom Clipping is set, the actual counting might not be deterministic, right? Someone might use a sketch or a deterministic count or some other unknown methodology in the future. For this reason, I don't see custom clipping as a counting methodology.
</blockquote></details>

My answer to your questions:
1. Yes that sounds right
2. The “universal max” here, do you mean the `maximum_frequency_per_user` in `measurement_spec`? This value will not be used for `DynamicClipCount` methodology. “deterministic” here if you mean noise, the Dynamic Clipping library adds noise to the input frequency histogram in order to find the optimized threshold. And currently the library is designed to take VID frequency histogram as input instead of sketch so that’s why it’s described as counting methodology. Do I understand it correctly?

Will let @jiayu-google and @chenweiw to add more clarity on the actual algorithm

jiayu-google (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-NpzdQye2nZfIBD1wWfi:b-azyhao) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

My answer to your questions:
1. Yes that sounds right
2. The “universal max” here, do you mean the `maximum_frequency_per_user` in `measurement_spec`? This value will not be used for `DynamicClipCount` methodology. “deterministic” here if you mean noise, the Dynamic Clipping library adds noise to the input frequency histogram in order to find the optimized threshold. And currently the library is designed to take VID frequency histogram as input instead of sketch so that’s why it’s described as counting methodology. Do I understand it correctly?

Will let @jiayu-google and @chenweiw to add more clarity on the actual algorithm
</blockquote></details>

Hi @stevenwarejones , here are some thoughts:
`DynamicClipCount` here means DeterministicCount + DynamicClip. (While `DeterministicCount` here means DeterministicCount + MCSpecifiedClip.) 

The clipping methodology is orthogonal to the counting methodology, so one option is to separate the messages of clipping and counting methodologies.

Albert's codes do not separate the clipping and counting methodologies. I think it's a reasonable approach for now, as we only support the deterministic counting --- we don't support LiquidLegions or other counting. If later we do not want support LiquidLegions counting, it makes sense to separate the messages of clipping and counting methodologies.

jiayu-google (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-NpziEVSAi_E6bRoC6v6:b-mv1hlk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

Hi @stevenwarejones , here are some thoughts:
`DynamicClipCount` here means DeterministicCount + DynamicClip. (While `DeterministicCount` here means DeterministicCount + MCSpecifiedClip.) 

The clipping methodology is orthogonal to the counting methodology, so one option is to separate the messages of clipping and counting methodologies.

Albert's codes do not separate the clipping and counting methodologies. I think it's a reasonable approach for now, as we only support the deterministic counting --- we don't support LiquidLegions or other counting. If later we do not want support LiquidLegions counting, it makes sense to separate the messages of clipping and counting methodologies.
</blockquote></details>

Sorry typo:
If later we do not want support LiquidLegions counting -> If later we DO want support LiquidLegions counting

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-NpzzJ78B8PWmRsPVzwa:b-dbgq0n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
<details><summary><i>Previously, jiayu-google wrote…</i></summary><blockquote>

Sorry typo:
If later we do not want support LiquidLegions counting -> If later we DO want support LiquidLegions counting
</blockquote></details>

@jiayu-google - I think we're roughly on the same page here. Two more thoughts:
1. I'm not a fan of conflating two different orthogonal methodologies. Can we have two fields: `counting_methodology` and `privacy_mechanisms` (that would include maximum clipping value)
2. For MCSpecifiedClip, where is this specified in the Measurement.Result? I don't think we support this today. Assuming, we don't support it today, it could also be under `privacy_mechanisms`

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-Nq-E4hq00hozD6X9k0c:b9w189c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*

@jiayu-google - I think we're roughly on the same page here. Two more thoughts:
1. I'm not a fan of conflating two different orthogonal methodologies. Can we have two fields: `counting_methodology` and `privacy_mechanisms` (that would include maximum clipping value)
2. For MCSpecifiedClip, where is this specified in the Measurement.Result? I don't think we support this today. Assuming, we don't support it today, it could also be under `privacy_mechanisms`
</blockquote></details>

@stevenwarejones Thanks for clarifying. I understand your point and agree. Dynamic Clipping is not a counting methodology.

Maybe we add a field to the existing counting mechanisms where dynamic clipping could apply (e.g. DeterministicCount, but perhaps in the future LiquidLegions) for EDPs to provide the clipped frequency value?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-Nq-VWoD9bicjBGnUQw5:b-3zkcl9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*

@stevenwarejones Thanks for clarifying. I understand your point and agree. Dynamic Clipping is not a counting methodology.

Maybe we add a field to the existing counting mechanisms where dynamic clipping could apply (e.g. DeterministicCount, but perhaps in the future LiquidLegions) for EDPs to provide the clipped frequency value?
</blockquote></details>

i'd be okay with adding it to the existing field for `DeterministicCount` - its not perfect but good enough for me

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @chenweiw, @iverson52000, @kungfucraig, @Marco-Premier, @riemanli, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 93 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq00ZV6EYm4_78RN4rw:-Nq00ZV6EYm4_78RN4rx:bkt3xk5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L93)):*
> ```proto
>   oneof privacy_mechanism {
>     // maximum_frequency_per_user passed in measurement_spec.
>     int32 mc_specified_cap = 1;
> ```

Let's not copy the field in more than on place. Just have `dynamic_clipping_maximum_frequency_per_user` or something and indicate that it's only specified when dynamic clipping is used. We can always wrap in a oneof later if we need more methodologies for computing maximum frequency.



_Code quote:_
```proto
int32 mc_specified_cap = 1;
```

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 0 of 4 files reviewed, 2 unresolved discussions (waiting on @chenweiw, @iverson52000, @kungfucraig, @Marco-Premier, @riemanli, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-Nq00yiZ1lN8O5DgA-H5:b93ax7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*

i'd be okay with adding it to the existing field for `DeterministicCount` - its not perfect but good enough for me
</blockquote></details>

I'd personally be fine with a `DynamicClippedDeterministicCount` methodology, i.e. just making it more specific. The maximum frequency does affect the counting, so I don't think it's really conflating concepts.

chenweiw (approved):
> Reviewed 4 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @iverson52000, @kungfucraig, @Marco-Premier, @riemanli, and @stevenwarejones)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @chenweiw, @iverson52000, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq2nz61BSRy5LMTlot4:b-d16da9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
> ```proto
>     // Optimized maximum frequency per user value calculated by the
>     // DataProvider.
>     int32 dynamic_clip_cap = 2;
> ```

why do we need a one of for this? What will the kingdom will do by knowing if its a mc_specified or dynamic?

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq3cy6E1VKcSOkYjZC3:b-ofv3g3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*

why do we need a one of for this? What will the kingdom will do by knowing if its a mc_specified or dynamic?
</blockquote></details>

Will take off the oneof following Sanjay's feedback. The actual `maximum_frequency_per_user` being used is for variance/bias calculation and the reporting server also needs to know about it. @jiayu-google @chenweiw Is that correct?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 238 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NputLm64JQ0-yGj1pLd:-Nq3e4tz3_WovJUNhf6C:b-cut6zx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/abc6da4789f6032fa1aff4e2e58d4f6aa796c6b7/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L238)):*
> I'd personally be fine with a `DynamicClippedDeterministicCount` methodology

Don't have a strong opinion on this but I notice the `Population` type also uses `DeterministicCount` methodology. Does `DeterministicCount` mean the same for `Population` and `Impression`?

https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L285

SanjayVas (commented):
> Reviewed 1 of 4 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @chenweiw, @iverson52000, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq3m-PH8hWhefG85J1n:bmznncw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

Will take off the oneof following Sanjay's feedback. The actual `maximum_frequency_per_user` being used is for variance/bias calculation and the reporting server also needs to know about it. @jiayu-google @chenweiw Is that correct?
</blockquote></details>

The Reporting system will know it for the static case as it creates the Measurement requests, including the MeasurementSpec. It only needs it from the EDP for the dynamic case.

Note also that the cap should also be the maximum key of the map in the returned result, so technically the EDP is already returning the cap. It's just that it's unclear whether the EDP use dynamic clipping without specifying so in the methodology.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq3noXh8lEmij_lo35j:b76m3gc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
> Note also that the cap should also be the maximum key of the map in the returned result

Actually right now for impression, `Measurement.Result` only has the noised impression value instead of the capped frequency histogram so EDP doesn't return the actual cap being used

https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L215

chenweiw (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @iverson52000, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq3sOoK7ox2q8oFEhK4:b-lzxf2y) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

> Note also that the cap should also be the maximum key of the map in the returned result

Actually right now for impression, `Measurement.Result` only has the noised impression value instead of the capped frequency histogram so EDP doesn't return the actual cap being used

https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/main/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L215
</blockquote></details>

@iverson52000 If "maximum_frequency_per_user" is the cap that dynamic clipping finally produced, then yes -- it is used for calculating the clipping bias and noise variance.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @chenweiw, @iverson52000, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq3yd2c6Oww7_llgbN7:bk6gtnb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
> Actually right now for impression, `Measurement.Result` only has the noised impression value instead of the capped frequency histogram so EDP doesn't return the actual cap being used

Ah sorry, I was mixing up R/F and Impression.

kungfucraig (changes requested):
> Reviewed 1 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 3 unresolved discussions (waiting on @chenweiw, @iverson52000, @jiayu-google, @Marco-Premier, and @riemanli)

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @chenweiw, @jiayu-google, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq97ZuvEp_17qCFunDf:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*

> Actually right now for impression, `Measurement.Result` only has the noised impression value instead of the capped frequency histogram so EDP doesn't return the actual cap being used

Ah sorry, I was mixing up R/F and Impression.
</blockquote></details>

Done.

stevenwarejones (changes requested):
> Reviewed 2 of 4 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @chenweiw, @jiayu-google, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nq9yFLR6ads04AruYfT:b-8pqltv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

Does it always have to ben an optimized value? What if an MC specifies a clipped value higher than what an EDP supports? Should the EDP reject the measurement, or instead just fulfill it with the value it supports? If the latter, than I think calling this `custom...` better captures multiple motivations for populating this field.

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-NqE5TuB9SvEXpQUkdTs:bob5enj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*

Does it always have to ben an optimized value? What if an MC specifies a clipped value higher than what an EDP supports? Should the EDP reject the measurement, or instead just fulfill it with the value it supports? If the latter, than I think calling this `custom...` better captures multiple motivations for populating this field.
</blockquote></details>

This PR is to add the support of dynamic clipping impression based on the dynamic clipping library which will come with the optimized frequency cap value(good impression accuracy and dp guarantee under ACDP and Gaussian noise). I don’t know if right now we are supporting EDP to choose a custom frequency cap value that it supports? Or it sounds like the `CustomDirectMethodology` we already have? cc @kungfucraig @jiayu-google

stevenwarejones (changes requested):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-NqZ7e823rqq1tLqzk8z:b84yvw0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

This PR is to add the support of dynamic clipping impression based on the dynamic clipping library which will come with the optimized frequency cap value(good impression accuracy and dp guarantee under ACDP and Gaussian noise). I don’t know if right now we are supporting EDP to choose a custom frequency cap value that it supports? Or it sounds like the `CustomDirectMethodology` we already have? cc @kungfucraig @jiayu-google
</blockquote></details>

@kungfucraig @jiayu-google - I'm not sure we need a separate field for optimized vs custom. I'd prefer to just call this custom. Is the math on the reporting server side going to be any different in each case?

SanjayVas (commented):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-NqZUMbu1DGploW-fEqD:bcc9ii0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
> I'm not sure we need a separate field for optimized vs custom

We do if we explicitly don't want to support the "custom" case. My understanding is that we want only two options at the moment: using dynamic clipping as described by the library or use the static value provided in the MeasurmentSpec. @kungfucraig can correct me if my understanding of that is incorrect.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-NqcmqglAUHRughWD6hM:b-ctbzu7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*

> I'm not sure we need a separate field for optimized vs custom

We do if we explicitly don't want to support the "custom" case. My understanding is that we want only two options at the moment: using dynamic clipping as described by the library or use the static value provided in the MeasurmentSpec. @kungfucraig can correct me if my understanding of that is incorrect.
</blockquote></details>

the alternative is that an EDP will need to reject the requisition.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nqcn-faDcX_kHQrisYF:bvhfjvk) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*

the alternative is that an EDP will need to reject the requisition.
</blockquote></details>

also - optimized is quite vague - optimized for what the legal team says is the right value? or optimized based on a certain halo algorithm?

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, @riemanli, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-Nqn5TtqCYIxGDBVSC5y:bdeyudc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*

also - optimized is quite vague - optimized for what the legal team says is the right value? or optimized based on a certain halo algorithm?
</blockquote></details>

Hi @stevenwarejones . Thanks for all the constructive feedback! After some internal discussion, we agree that the changes here should be a custom behavior for EDP instead of enforcing EDP to follow a specific algorithm. Update the field name to `custom_maximum_frequency_per_user` and the PR tile and description to reflect that. Please review. Although a custom cap might have some other implications in terms of the variance/bias calculation which requires more analysis by the data scientists.

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r4, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @chenweiw, @iverson52000, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NqnKi8tE3VXBglv3ZPZ:-NqnKi8u3i_eckekcpfi:ba3woi3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/53ec27aa82aeb392f1494e737ef1ba705d5e5704/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*
> ```proto
>     // Maximum frequency per user that would be included in this measurement.
>     //
>     // If the custom_maximum_frequency_per_user in the DeterministicCount
> ```

Need to update this to match the changes in methodology messages.



_Code quote:_
```proto
    // If the custom_maximum_frequency_per_user in the DeterministicCount
```

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 2 unresolved discussions (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto` line 97 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-Nq2nz605Yv-BJABJzvi:-NrbBiaS4-QkbpqmcRaT:bse9dvx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b1ff1b52d6316020034718ce8aa34e54e9029ec4/src/main/proto/wfa/measurement/api/v2alpha/direct_computation.proto#L97)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

Hi @stevenwarejones . Thanks for all the constructive feedback! After some internal discussion, we agree that the changes here should be a custom behavior for EDP instead of enforcing EDP to follow a specific algorithm. Update the field name to `custom_maximum_frequency_per_user` and the PR tile and description to reflect that. Please review. Although a custom cap might have some other implications in terms of the variance/bias calculation which requires more analysis by the data scientists.
</blockquote></details>

Hi @stevenwarejones. Did you have time to review the subsequent changes?

stevenwarejones (approved):
> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw, @iverson52000, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

iverson52000 (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: all files reviewed, 1 unresolved discussion (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, @riemanli, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NqnKi8tE3VXBglv3ZPZ:-NrkruuO76qws6QEPUWY:byz3jgy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/53ec27aa82aeb392f1494e737ef1ba705d5e5704/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*

Need to update this to match the changes in methodology messages.
</blockquote></details>

Hi @SanjayVas. What should I update here?

SanjayVas (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @chenweiw, @jiayu-google, @kungfucraig, @Marco-Premier, and @riemanli)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto` line 97 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-NqnKi8tE3VXBglv3ZPZ:-Nrl1QNMET3G2UPqQSZO:b-ivsjvt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/53ec27aa82aeb392f1494e737ef1ba705d5e5704/src/main/proto/wfa/measurement/api/v2alpha/measurement_spec.proto#L97)):*
<details><summary><i>Previously, iverson52000 (Albert Hsu) wrote…</i></summary><blockquote>

Hi @SanjayVas. What should I update here?
</blockquote></details>

Looks like you've done the update. It was previously still referencing DynamicClipCount.

kungfucraig (approved):
> Reviewed 2 of 4 files at r4, 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @chenweiw, @jiayu-google, @Marco-Premier, and @riemanli)

chenweiw (approved):
:lgtm:

> Reviewed 2 of 2 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/197#-:-NrlanLnELg18p5kQlbW:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @jiayu-google, @Marco-Premier, and @riemanli)

Issue #220, opened 2024-12-03
Set Data Availability per Model Line


stevenwarejones:
Currently, data availability is set once per Event Group (open-ended) and once per Data Provider (closed-ended).

However, when additional model lines are released, it is highly likely that EDPs will label their data with overlapping model lines that do not provide complete coverage to either model line. Thus, EDPs need to report which model lines are being used for which event groups so that measurements can be correctly constructed.

kungfucraig:
Currently Origin is updating the model line as they upgrade since these introduce new dimensions and significantly update the modeling techniques used. Depending on the steady state rollout plan this could continue to happen, or they could add model releases, in which case we'd only expect a single long-lived model line for production usage.

If we're only talking about model releases for a single model line these per-model line availability fields are not strictly necessary. 

However, for completeness we should have them since even if we have long-lived "prod" model lines, there will be experimental model lines that can take advantage of this capability. Supporting multiple markets with a single Kingdom will require this too.

I'm not convinced we need per model line availability at the Event Group level. I made a comment on the PR to this effect.

SanjayVas:
Resolved by #226, which introduces data availability by `ModelLine` for `DataProvider` resources. All `EventGroup`s are expected to support all valid `ModelLine`s for the intersection of the data availability interval of the `EventGroup` and the data availability for the `ModelLine`.

PR #221, opened 2024-12-03
add availability per model line for DataProvider and EventGroup

stevenwarejones:
Addresses https://github.com/world-federation-of-advertisers/cross-media-measurement-api/issues/220

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto
@@ -68,9 +68,9 @@ message DataProvider {
     (google.api.field_behavior) = UNORDERED_LIST
   ];
 
-  // Interval for when data is guaranteed to be available for a `Requisition`.
-  // If this field is specified, both `start_time` and `end_time` must be set.
-  google.type.Interval data_availability_interval = 7;
+  // Deprecated: Use `data_availability` field except for legacy resoures
+  // where it is not set.
+  google.type.Interval data_availability_interval = 7 [deprecated = true];
 
   // Capabilities of a `DataProvider`.
   message Capabilities {
@@ -82,4 +82,17 @@ message DataProvider {
   // This indicates to the CMMS what the `DataProvider` is capable of at
   // `Measurement` creation time.
   Capabilities capabilities = 8;
+
+  message DataAvailability {
+    // Resource name of the ModelLine's availability.
+    string model_line = 1 [
+      (google.api.resource_reference).type = "halo.wfanet.org/ModelLine",
+      (google.api.field_behavior) = IMMUTABLE
+    ];
+
+    // Interval for when data is guaranteed to be available for a `Requisition`.
+    // If this field is specified, both `start_time` and `end_time` must be set.
+    google.type.Interval interval = 2;
+  }
+  repeated DataAvailability data_availability = 9;
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
@@ -128,17 +128,31 @@ message EventGroup {
   // Event Group state.
   State state = 9 [(google.api.field_behavior) = OUTPUT_ONLY];
 
-  // Interval for when data for this `EventGroup` is guaranteed to be available
-  // for a `Requisition`.
-  //
-  // The parent `data_availability_interval` defines the upper and lower bounds
-  // for when data is guaranteed to be available.
-  //
-  // This field will be required in a future release. Until then, if this is not
-  // set, data is not guaranteed to be available.
-  //
-  // If this is set, then `start_time` is required and the data is guaranteed to
-  // be available for the interval that defines the overlap between this and the
-  // parent `data_availability_interval`.
-  google.type.Interval data_availability_interval = 11;
+  // Deprecated: Use `data_availability` field except for legacy resoures
+  // where it is not set.
+  google.type.Interval data_availability_interval = 11 [deprecated = true];
+
+  message DataAvailability {
+    // Resource name of the ModelLine's availability.
+    string model_line = 1 [
+      (google.api.resource_reference).type = "halo.wfanet.org/ModelLine",
+      (google.api.field_behavior) = IMMUTABLE
+    ];
+
+    // Interval for when data for this `EventGroup` is guaranteed to be available
+    // for a `Requisition`.
+    //
+    // The parent `DataProvider` `data_provider.data_availability.interval`
+    // defines the upper and lower bounds for when data is guaranteed to be
+    // available.
+    //
+    // This field will be required in a future release. Until then, if this is not
+    // set, data is not guaranteed to be available.
+    //
+    // If this is set, then `start_time` is required and the data is guaranteed to
+    // be available for the interval that defines the overlap between this and the
+    // parent `DataProvider`s `data_availability.interval`.
+    google.type.Interval interval = 2;
+  }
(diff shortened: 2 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221)

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 135 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221#-ODCVSJgDt3hlQCkAh50:-ODCf_CkCe89bT7HpKr-:b-9rkau5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/afc8e33a7f655e36e0169a2bc71e201889f27962/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L135)):*

Maybe we don't need per model line availability at the Event Group level since the EG is always constrained to be a subset of the DP availability. 

At least, I can't think of a case where it would help and having this level of granularity seriously complicates things.

Maybe you have a case in mind?
</blockquote></details>

It is completely needed if EDPs don't label all campaigns with all available VID Models on a given day. In that case, they can't even populate the DataProvider fields at all. Maybe the DataProvider availability should be optional in that case.

I see this scenario playing out if EDPs always label a campaign with whatever model line they first started labeling the campaign.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 135 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221#-ODCVSJgDt3hlQCkAh50:-ODD7jkkCNzp7qpa99Dj:bm2tv5i) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/afc8e33a7f655e36e0169a2bc71e201889f27962/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L135)):*

It is completely needed if EDPs don't label all campaigns with all available VID Models on a given day. In that case, they can't even populate the DataProvider fields at all. Maybe the DataProvider availability should be optional in that case.

I see this scenario playing out if EDPs always label a campaign with whatever model line they first started labeling the campaign.
</blockquote></details>

That's pretty hypothetical and adds a lot of additional complexity. 

It could always be added in the future if/when we encounter the scenario you mention. We could even oneof it with the global data availability field and maintain backward compatibility.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 135 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221#-ODCVSJgDt3hlQCkAh50:-ODJCLMV22mU-lvvPvVr:b-sphs0g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/afc8e33a7f655e36e0169a2bc71e201889f27962/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L135)):*

That's pretty hypothetical and adds a lot of additional complexity. 

It could always be added in the future if/when we encounter the scenario you mention. We could even oneof it with the global data availability field and maintain backward compatibility.
</blockquote></details>

Its very real. We've talked about that scenario. My takeaway is that each market (or possibly globally) will either go in the direction of needing it on Event Groups or on the Data Provider but probably not both (per market).

Maybe we should wait until there's direction on that.

Or it could evolve with the DataProvider one being needed and then the market evolves to needing it on the Event Group level.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 135 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221#-ODCVSJgDt3hlQCkAh50:-ODJxZK75SWT16Je4kGg:b-qjut2g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/afc8e33a7f655e36e0169a2bc71e201889f27962/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L135)):*

Its very real. We've talked about that scenario. My takeaway is that each market (or possibly globally) will either go in the direction of needing it on Event Groups or on the Data Provider but probably not both (per market).

Maybe we should wait until there's direction on that.

Or it could evolve with the DataProvider one being needed and then the market evolves to needing it on the Event Group level.
</blockquote></details>

So we did talk about the scenario of continuing to label a given EG with the same model line for the life time of the EG, but I don't think that's the right way to approach labeling because in general that will not allow you to meaningfully report across EGs. 

However, even if it were desirable I'm not convinced that it requires this field. 

I see some options for moving forward. In my order of preference:
1) Hold on all of this until we have some guidance from the model provider about how they want to manage VID model life times.
2) Write a doc that describes the cases that require this and demonstrate the need. I started to work an example and I don't think these fields are necessary, but I didn't have time to work it all the way through. And there may be other cases.
3) Drop the fields from the PR and we can deal with adding them when we have an identified need.

stevenwarejones (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221)* status: 0 of 2 files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 135 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/221#-ODCVSJgDt3hlQCkAh50:-ODRtlqa6Ul7LBafa443:b-6nhtvg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/afc8e33a7f655e36e0169a2bc71e201889f27962/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L135)):*

So we did talk about the scenario of continuing to label a given EG with the same model line for the life time of the EG, but I don't think that's the right way to approach labeling because in general that will not allow you to meaningfully report across EGs. 

However, even if it were desirable I'm not convinced that it requires this field. 

I see some options for moving forward. In my order of preference:
1) Hold on all of this until we have some guidance from the model provider about how they want to manage VID model life times.
2) Write a doc that describes the cases that require this and demonstrate the need. I started to work an example and I don't think these fields are necessary, but I didn't have time to work it all the way through. And there may be other cases.
3) Drop the fields from the PR and we can deal with adding them when we have an identified need.
</blockquote></details>

I prefer option 3. Only have the dataprovider fields. The current `main` behavior is hard to reason what to do once you add a new modelline. Of course, there are no model lines to report on.

PR #225, opened 2025-01-31
feat: Add EventGroup metadata

kungfucraig:


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)

SanjayVas (changes requested):
> Reviewed 1 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 1 of 5 files reviewed, 7 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 59 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm2hJWACV5RHQkVIyp:-OIm2hJWACV5RHQkVIyq:undefined) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L59)):*
> ```proto
>   // ID referencing the `EventGroup` in an external system, provided by the
>   // `DataProvider`. This value should be used to help the `DataProvider`
>   // synchornize their metadata with the EventGroupsService. See
> ```





_Suggestion:_
```proto
synchronize
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 60 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm2NP39pQPm2GQmhs6:-OIm2NP39pQPm2GQmhs7:bpxc446) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L60)):*
> ```proto
>   // `DataProvider`. This value should be used to help the `DataProvider`
>   // synchornize their metadata with the EventGroupsService. See
>   // `EventGroupMetadata` if you wish you provide end-user visible identifiers.
> ```

This is also end-user visible. Really the only difference between this and the fields in metadata is that this optional and has no prescribed mapping other than being *some* identifier in the DataProvider's system.



_Code quote:_
```proto
`EventGroupMetadata` if you wish you provide end-user visible identifiers
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 61 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm4hBv-k43BNImBuKb:-OIm4hBv-k43BNImBuKc:b-bv4eo6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L61)):*
> ```proto
>   // synchornize their metadata with the EventGroupsService. See
>   // `EventGroupMetadata` if you wish you provide end-user visible identifiers.
>   string event_group_reference_id = 5 [(google.api.field_behavior) = OPTIONAL];
> ```

If you're starting to use this field behavior, you need to do it consistently across the API. This API was designed [before field behavior was mandated on all fields](https://google.aip.dev/203#history), so it is following the old convention that anything without the REQUIRED is optional.



_Code quote:_
```proto
(google.api.field_behavior) = OPTIONAL
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm4GDXE5dmG4FPkSUJ:-OIm4GDXE5dmG4FPkSUK:b-g29s3z) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L83)):*
> ```proto
>   // The `EventTemplate`s that events associated with this `EventGroup` conform
>   // to.
>   repeated EventTemplate event_templates = 7
> ```

This is not deprecated and must still be specified. I think you're confusing this with the metadata descriptor.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 107 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm3ZQE8nMG88zKm5v2:-OIm3ZQE8nMG88zKm5v3:b-hindbu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L107)):*
> ```proto
>   // The version of the `Metadata` message type must match the API version used
>   // to last update this resource.
>   EncryptedMessage encrypted_metadata = 10 [deprecated = true];
> ```

The deprecation needs to be documented, including what field to use instead and under what conditions this still needs to be written/read.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 111 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm3FoF7B6H3ctigRri:-OIm3FoF7B6H3ctigRrj:b4xgvo5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L111)):*
> ```proto
>   // Encrypted serialized `Metadata`.
>   //
>   // Deprecated: Use `encrypted_metadata` instead.
> ```

Note that there's now a deprecated field pointing to another deprecated field. We should clean up all the existing deprecated fields and go to 0.6.0 first before deprecating any other fields.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 140 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm3vvd5SmzsE5DUAsX:-OIm3vvd5SmzsE5DUAsY:b-25mato) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L140)):*
> ```proto
> 
>   // Interval for when events for this `EventGroup` are available.
> ```

Don't drop the guarantee part. That's important, as it's what makes it clear that it is an error for the DataProvider to not have available data for this range.



_Code quote:_
```proto
are available.
```

SanjayVas (changes requested):
> Reviewed 2 of 4 files at r2, 3 of 3 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 3 of 5 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm4GDXE5dmG4FPkSUJ:-OIrKQAv2g9ZY20Za4cb:bk8rxwq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L83)):*

Updated the comment and added media_types.
</blockquote></details>

I don't think the comment is accurate, or at least not something we've discussed. Is the idea to have a config for the Kingdom which indicates the possible templates (perhaps by specifying an event message type)? I understand the part about deriving which templates based on media type after that.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIrLKIf1VNTAzWA3TUj:-OIrLKIf1VNTAzWA3TUk:brurow3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ebf0628e1141d2c3f3b3c42f6374d659e7dcc769/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L41)):*
> ```proto
>     // Metadata the pertains to a campaign that is entailed by the EventGroup.
>     CampaignMetadata campaign_metadata = 1
>         [(google.api.field_behavior) = OPTIONAL];
> ```

Shouldn't this be required? Meaning if you select ad_metadata, you must specify campaign_metadata for that ad?

Or taking a step back, what do you expect to be adding to campaign metadata outside of ad metadata? My instinct was to just call this AdCampaignMetadata and drop a layer.



_Code quote:_
```proto
OPTIONAL
```

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 2 of 5 files reviewed, 3 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIrLKIf1VNTAzWA3TUj:-OIrVrXU5wXXd0nM2zE5:bt10107) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ebf0628e1141d2c3f3b3c42f6374d659e7dcc769/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L41)):*

Shouldn't this be required? Meaning if you select ad_metadata, you must specify campaign_metadata for that ad?

Or taking a step back, what do you expect to be adding to campaign metadata outside of ad metadata? My instinct was to just call this AdCampaignMetadata and drop a layer.
</blockquote></details>

I'm working under the assumption we'll eventually have things like CreativeMetadata, PlacementMetadata, etc.

SanjayVas (commented):
> Reviewed 3 of 3 files at r4.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 83 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm4GDXE5dmG4FPkSUJ:-OIvhw6Y80SxqR2v-2kK:bfltzmv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L83)):*

I don't think the comment is accurate, or at least not something we've discussed. Is the idea to have a config for the Kingdom which indicates the possible templates (perhaps by specifying an event message type)? I understand the part about deriving which templates based on media type after that.
</blockquote></details>

Discussed offline. The set of possible event templates for a market only needs to be configured/read at the Reporting system level, but is still describing the set of valid templates for a single Kingdom/market. This means that if there are multiple Reporting system instances connected to a given Kingdom, all of them will need the same set of templates configured. That said, this configuration may not be supported in the future if Reporting and the Kingdom are merged.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 111 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIm3FoF7B6H3ctigRri:-OIviX9Y9xVUMP8oXfnJ:blqjjo1) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/249026fd26c006a7e4a60e11a14f593c0bc6a263/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L111)):*

Let's discuss this offline.
</blockquote></details>

Discussed offline. Not deprecating any fields that are in use now, as there is a period of time where the existing fields will still need to be written even if the new ones are used. Regardless, the earliest the existing fields could be marked as deprecated is 0.6.0 and the earliest they could be removed is 0.7.0

kungfucraig (commented):
> Reviewed 2 of 3 files at r4, 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIrLKIf1VNTAzWA3TUj:-OJFwGav-Iqb5tPZsdtN:ba8rctt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ebf0628e1141d2c3f3b3c42f6374d659e7dcc769/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L41)):*

Doesn't that still mean that campaign_metadata should be required? Or is it something that you think would be inherited once we have hierarchical EventGroups? Even so, it should probably be required for now until that design is sorted out.
</blockquote></details>

I anticipate that this may get wrapped in a oneof eventually, , but making it required for now is not at odds with that.

SanjayVas (approved):
> Reviewed 2 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @stevenwarejones)

stevenwarejones (changes requested):
> Reviewed 2 of 3 files at r3, 2 of 3 files at r4, 1 of 2 files at r5, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIrLKIf1VNTAzWA3TUj:-OIw-qwW9kpmBCOHhHKi:bgb0mkr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ebf0628e1141d2c3f3b3c42f6374d659e7dcc769/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L41)):*

Doesn't that still mean that campaign_metadata should be required? Or is it something that you think would be inherited once we have hierarchical EventGroups? Even so, it should probably be required for now until that design is sorted out.
</blockquote></details>

I can't think of a time when campaignMetaData should not be required unless we are extending beyond impression data.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 74 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtfaGf9F6WPHG752Py:-OJtfaGf9F6WPHG752Pz:b-e5p2h3) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L74)):*
> ```proto
>   //
>   // TODO(@kungfucraig): Make this REQUIRED once it has been adopted.
>   EventGroupMetadata event_group_metadata = 14;
> ```

where is this specified?

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgIwPAxZSbHbe-h3v:-OJtgIwPAxZSbHbe-h3w:btk75zt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L36)):*
> ```proto
>       string campaign_name = 2 [(google.api.field_behavior) = REQUIRED];
>       // Event Data Provider identifiers for the Campaign
>       repeated string data_provider_identifiers = 3 [
> ```

what is the scenario where this needs to be repeated? What about repeating it within the event group itself (see above)
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 74 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgU4i66JDZ0huBU7J:-OJtgU4i66JDZ0huBU7K:bth23os) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L74)):*
> ```proto
>   //
>   // TODO(@kungfucraig): Make this REQUIRED once it has been adopted.
>   EventGroupMetadata event_group_metadata = 14;
> ```

wdyt about making this a repeated

kungfucraig (commented):
> Reviewed 3 of 3 files at r7, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 74 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtfaGf9F6WPHG752Py:-OJuRk1X2qsZMmbZGWHp:b-fg8n44) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L74)):*

where is this specified?
</blockquote></details>

In the file below.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 74 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgU4i66JDZ0huBU7J:-OJuRs_v6J1GKmkLtPn4:b-ovey7n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L74)):*

wdyt about making this a repeated
</blockquote></details>

Let's create that kind of substructure inside of EventGroupMetadata itself.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 41 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OIrLKIf1VNTAzWA3TUj:-OJuT8erEOaL_mwnfv1r:bssh3cr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ebf0628e1141d2c3f3b3c42f6374d659e7dcc769/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L41)):*

I can't think of a time when campaignMetaData should not be required unless we are extending beyond impression data.
</blockquote></details>

It depends. If we have Event Group hierarchies we may find that the campaign metadata for a creative can be inferred by the hierarchy and does not need to be duplicated. At any rate making this required for now keeps our options open.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgIwPAxZSbHbe-h3v:-OJuTonvBGDU74576YfI:bm4x9xo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L36)):*

what is the scenario where this needs to be repeated? What about repeating it within the event group itself (see above)
</blockquote></details>

In the case that the data provider mapping is somehow onto. I don't think we should preclude that possibility.

stevenwarejones (changes requested):
> Reviewed 1 of 4 files at r8, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 74 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgU4i66JDZ0huBU7J:-OJzRTyX9lX7Rn7Ed1Kl:b3vw7k8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L74)):*

Let's create that kind of substructure inside of EventGroupMetadata itself.
</blockquote></details>

if not here, then see my other comment about adding a different repeated
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 42 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJzRPP9Bv6M85yK4oJQ:-OJzRPP9Bv6M85yK4oJR:b-ou0h5r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f75273079876909bd2ca72538512f60fc57c3b6e/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L42)):*
> ```proto
>     }
>     // Metadata the pertains to a campaign that is entailed by the EventGroup.
>     CampaignMetadata campaign_metadata = 1
> ```

i think this should be a repeated

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 4 of 5 files reviewed, 3 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 42 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJzRPP9Bv6M85yK4oJQ:-OJzY4h23VhkfrRAxn5p:bc83ppq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f75273079876909bd2ca72538512f60fc57c3b6e/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L42)):*

i think this should be a repeated
</blockquote></details>

The goal of this exercise is to replicate the Origin metadata and add campaign identifiers (which are a known gap). 

It's worth having the broader discussion, and I'm not sure I agree with repeating this structure. Fortunately, the structure is extendable and once we learn more we can extend it.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 4 of 5 files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgIwPAxZSbHbe-h3v:-OK1Y8l-6hzM9YX1FWL9:b6qri5r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L36)):*

In the case that the data provider mapping is somehow onto. I don't think we should preclude that possibility.
</blockquote></details>

maybe we should then be adding an advertiser_taxonomy_type field or something to denote what taxaonomy these repeated fields correspond to. I'd argue if they are different campaign IDs then we should discourage them repeating them here and instead encourage repeated CampaignMEtadata
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 42 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJzRPP9Bv6M85yK4oJQ:-OK1Y-jE37raAEbx1Dj3:b-krt1xg) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f75273079876909bd2ca72538512f60fc57c3b6e/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L42)):*

i easily see the scenario where a single event group maps to multiple campaigns.
</blockquote></details>

I'm pretty sure we're going to need this to be repeated in the future but assuming this is forward compatible with moving to repeated, them I'm okay with merging for now.

kungfucraig (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: 4 of 5 files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgIwPAxZSbHbe-h3v:-OK2LEwXBv3oTgGle8XW:b-l3u2cu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L36)):*

maybe we should then be adding an advertiser_taxonomy_type field or something to denote what taxaonomy these repeated fields correspond to. I'd argue if they are different campaign IDs then we should discourage them repeating them here and instead encourage repeated CampaignMEtadata
</blockquote></details>

Let's replicate the existing Origin metadata so we have a trivial and clear migration path and then plan to evolve this over time. 

To that end I'm fine dropping this identifier field for now if that works for you.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 42 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJzRPP9Bv6M85yK4oJQ:-OK2L6y69k_mslCg9gLL:b5p66p2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f75273079876909bd2ca72538512f60fc57c3b6e/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L42)):*

I'm pretty sure we're going to need this to be repeated in the future but assuming this is forward compatible with moving to repeated, them I'm okay with merging for now.
</blockquote></details>

Sounds good. Let's move ahead with this for now.

kungfucraig (commented):
> Reviewed 4 of 4 files at r8, 1 of 1 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 36 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225#-OJtgIwPAxZSbHbe-h3v:-OL-mpK64cJO1P7e3W_E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/351afcc001ebe0a9d73653997917209cedc05064/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L36)):*

lets drop the identifier then. There's a few different scenarios that need to be covered...
1. each identifier is the save event type
2. they are mixed event types
</blockquote></details>

Done.

kungfucraig (commented):
> Reviewed 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

kungfucraig (commented):
> Reviewed 1 of 1 files at r11, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

kungfucraig (commented):
> Reviewed 2 of 2 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/225)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #226, opened 2025-02-01
feat: Add per ModelLine data availability to DataProvider

kungfucraig:

PR #237, opened 2025-05-08
docs: Clarify documentation of data availability fields

SanjayVas:
This attempts to resolve the ambiguity of how Requisitions should be fulfilled when any part of the interval falls outside of the effective data availability interval of an EventGroup. Specifically, that Requisitions should be fulfilled as if there were 0 events outside of that interval (as opposed to being refused).

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto
@@ -87,7 +87,7 @@ message DataProvider {
   // `Measurement` creation time.
   Capabilities capabilities = 8;
 
-  // Per `ModelLine` data availabiity
+  // An entry in the [data_availability_intervals][] map.
   message DataAvailabilityMapEntry {
     // Resource name of a `ModelLine`
     string key = 1 [
@@ -95,14 +95,24 @@ message DataProvider {
       (google.api.field_behavior) = REQUIRED
     ];
 
-    // Interval for when data is guaranteed to be available for a `Requisition`
-    // fulfillment. If this field is specified, both `start_time` and `end_time`
-    // must be set.
+    // Interval for when data is guaranteed to be available for a
+    // [Requisition][] fulfillment for the [ModelLine][] indicated by [key][].
+    //
+    // A [DataProvider][] must only specify availability for intervals
+    // when they are guaranteed to have the data available to fulfill any
+    // [Requisition][] within that interval. See
+    // [RequisitionSpec.EventGroupEntry.Value.collection_interval][].
+    //
+    // If this field is specified, both `start_time` and `end_time` must be set.
     google.type.Interval value = 2 [(google.api.field_behavior) = REQUIRED];
   }
-  // `DataAvailability` for each `ModelLine` that the `DataProvider` supports.
+  // A map of [ModelLine][] resource name to data availability interval.
+  //
   // TODO(@kungfucraig): Make this REQUIRED once it has been adopted, and no
-  // earlier than the 0.7 release.
+  // earlier than the 0.7 release. Once this is required, it may be an error to
+  // create a [Measurement][] unless the interval specified by
+  // [RequisitionSpec][] does not fall into the availability interval for
+  // [MeasurementSpec.model_line][].
   repeated DataAvailabilityMapEntry data_availability_intervals = 9
       [(google.api.field_behavior) = UNORDERED_LIST];
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
@@ -168,16 +168,16 @@ message EventGroup {
   // Event Group state.
   State state = 9 [(google.api.field_behavior) = OUTPUT_ONLY];
 
-  // Interval for when events for this `EventGroup` are available.
+  // Interval for which this [EventGroup][] has events.
   //
-  // For an Ad Campaign these are the start and end dates of the campaign.
+  // When this field is specified, [start_time][google.type.Interval.start_time]
+  // is required. If [end_time][google.type.Interval.end_time] is not specified,
+  // the interval is considered unbounded.
   //
-  // Events are available in the range specified here, but are only guaranteed
-  // to be available for reporting for a paticular `ModelLine` where this range
-  // intersects with the `data_availability_intervals` set forth on the
-  // `DataProvider` for the particular `ModelLine`
+  // If this [EventGroup][] represents an ad campaign, this is typically the
+  // start and end times of the campaign.
   //
   // TODO(@kungfucraig): Make this REQUIRED once it has been adopted. This will
-  // happen no sooner than the 0.6 relesae.
+  // happen no sooner than the 0.6 release.
   google.type.Interval data_availability_interval = 11;
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto
@@ -145,7 +145,10 @@ message Measurement {
   message Result {
     // A reach result.
     message Reach {
-      // Number of unique users exposed.
+      // Number of unique users (VIDs) for which there are impression events.
+      //
+      // This must be non-negative, meaning any negative value arising due to
+      // noise must be clamped to 0.
       int64 value = 1;
 
       // The mechanism used to generate noise during computation.
@@ -183,9 +186,13 @@ message Measurement {
     // A frequency result.
     message Frequency {
       // Map of frequency to reach ratio. For example, an entry
-      // {key: 4 value: 0.333} means that 33.3% of users were exposed exactly 4
-      // times, unless 4 is the largest key (maximum frequency) in which case it
-      // means that 33.3% of users were exposed at least 4 times.
+      // {key: 4 value: 0.333} means that 33.3% of users have exactly 4
+      // impression events, unless 4 is the largest key (maximum frequency) in
+      // which case it means that 33.3% of users have *at least* 4 impression
+      // events.
+      //
+      // Values must be non-negative, meaning any negative value arising due to
+      // noise must be clamped to 0.
       map<int64, double> relative_frequency_distribution = 1;
 
       // The mechanism used to generate noise during computation.
@@ -219,7 +226,10 @@ message Measurement {
 
     // An impression result.
     message Impression {
-      // Number of total impressions.
+      // Number of impression events.
+      //
+      // This must be non-negative, meaning any negative value arising due to
+      // noise must be clamped to 0.
       int64 value = 1;
 
       // The mechanism used to generate noise during computation.
@@ -245,6 +255,9 @@ message Measurement {
     // A watch duration result.
     message WatchDuration {
       // Total duration.
+      //
+      // This must be non-negative, meaning any negative value arising due to
+      // noise must be clamped to 0.
       google.protobuf.Duration value = 1;
 
       // The mechanism used to generate noise during computation.
@@ -270,6 +283,8 @@ message Measurement {
     // A population result.
     message Population {
       // The population value.
+      //
+      // This must be non-negative.
       int64 value = 1;
 
       // The computation methodology done by data provider. Required.
--- a/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto
@@ -238,14 +238,27 @@ message Requisition {
       // This implies that the `DataProvider` *should* be able to fulfill the
       // `Requisition`, but something went irrecoverably wrong. For example, the
       // `DataProvider` encountered corruption of the underlying data.
+      //
+      // This indicates a failure within the `DataProvider`'s fulfillment
+      // mechanism.
       UNFULFILLABLE = 4;
 
       // The `DataProvider` has declined to fulfill this `Requisition`
       // regardless of whether any of the other `Justification` conditions
       // apply.
+      //
+      // For example, a `DataProvider` policy that cannot be expressed within
+      // the API.
       DECLINED = 5;
     }
-    // Justification for refusing to fulfill this `Requisition`.
+    // Justification for refusing to fulfill this [Requisition][].
+    //
+    // Note that there is no [Justification][] for
+    // [collection_interval][RequisitionSpec.EventGroupEntry.Value.collection_interval]
+    // falling outside of [EventGroup.data_availability_interval][], as that is
+    // allowed behavior. See the documentation for
+    // [RequisitionSpec.EventGroupEntry.Value.collection_interval][] for more
+    // details.
     Justification justification = 1 [(google.api.field_behavior) = REQUIRED];
 
     // Human-readable string adding more context to the provided
--- a/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto
@@ -29,7 +29,11 @@ option go_package = "github.com/world-federation-of-advertisers/cross-media-meas
 
 // Duchy service for fulfilling `Requisition`s.
 service RequisitionFulfillment {
-  // Fulfills a `Requisition`.
+  // Fulfills a [Requisition][] using a protocol other than
+  // [Direct][ProtocolConfig.Direct].
+  //
+  // See [Requisitions.FulfillDirectRequisition][] for fulfilling a
+  // [Requisition][] using the [Direct][ProtocolConfig.Direct] protocol.
   rpc FulfillRequisition(stream FulfillRequisitionRequest)
       returns (FulfillRequisitionResponse) {}
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/requisition_spec.proto
@@ -45,9 +45,17 @@ message RequisitionSpec {
 
     // Value of an `EventGroupEntry`.
     message Value {
-      // Time interval over which the event data should be collected.
+      // Time interval over which event data is collected for fulfillment.
       //
-      // Both `start_time` and `end_time` must be specified.
+      // Both [start_time][google.type.Interval.start_time] and
+      // [end_time][google.type.Interval.end_time] must be specified.
+      //
+      // Only events in the intersection of this, the
+      // [DataProvider.data_availability_intervals][] entry for
+      // [MeasurementSpec.model_line][], and the
+      // [EventGroup.data_availability_interval][] for the [EventGroup][]
+      // matching [key][] need be included. Outside of that intersection, the
+      // [EventGroup][] may be considered to have 0 events.
       google.type.Interval collection_interval = 1
           [(google.api.field_behavior) = REQUIRED];
 
--- a/src/main/proto/wfa/measurement/api/v2alpha/requisitions_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/requisitions_service.proto
@@ -36,16 +36,22 @@ service Requisitions {
     option (google.api.method_signature) = "parent";
   }
 
-  // Transitions a `Requisition` to the `REFUSED` state.
+  // Transitions a [Requisition][] to the [REFUSED][Requisition.State.REFUSED]
+  // state.
   //
-  // This is a terminal state for the `Requisition` and all computations that
-  // rely on the `Requisition` will fail. Consequently, this should only be used
-  // for permanent failures and not transient errors.
+  // This is a terminal state for the [Requisition][], resulting in the
+  // permanent failure of the parent [Measurement][]. Consequently, this should
+  // only be called for permanent failures and not transient errors.
   //
-  // This is a state transition method (see https://aip.dev/216).
+  // This is a [state transition method](https://aip.dev/216).
   rpc RefuseRequisition(RefuseRequisitionRequest) returns (Requisition);
 
-  // Fulfills a `Requisition` directly without calling a `Duchy`.
+  // Fulfills a [Requisition][] using the [Direct][ProtocolConfig.Direct]
+  // protocol.
+  //
+  // This may only be called if [Requisition.protocol_config][] indicates that
+  // the protocol may be used. See the [RequisitionFulfillment][] service to
+  // fulfill using a different protocol.
   rpc FulfillDirectRequisition(FulfillDirectRequisitionRequest)
       returns (FulfillDirectRequisitionResponse);
 }


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)

kungfucraig (changes requested):
> Reviewed 2 of 4 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: 2 of 4 files reviewed, 6 unresolved discussions

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqCSuB1cPHjfGpYUeq:-OPqCSuB1cPHjfGpYUer:b-btncld) (no related file):*
It would be useful to add some additional context to requisitions_service.proto referencing the resource documentation where necessary. The service definitions are a like entry point for people.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 171 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqDPvj26fs9c5eo_j9:-OPqDPvj26fs9c5eo_jA:b-o341i7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L171)):*
> ```proto
> 
>   // Interval for which this [EventGroup][] is known to have events, subject to
> ```

Adding the "subject to" doesn't provide clarity. This is the availability of events for the Event Group. How it relates to DP data availability is a function of requisition fulfillment.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 181 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPpt2jD4-a9jSgwAVsl:-OPpt2jD4-a9jSgwAVsm:bvvt8ww) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L181)):*
> ```proto
>   // for times outside of this interval.
>   //
>   // If this field is specified, [start_time][google.type.Interval.start_time]
> ```

s/If/When so that the documentation will age better when we make it required.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 189 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqCkfE7hBVQdiW19jm:-OPqCkfE7hBVQdiW19jn:bnzoxnv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L189)):*
> ```proto
>       // impression events, unless 4 is the largest key (maximum frequency) in
>       // which case it means that 33.3% of users have *at least* 4 impression
>       // events.
> ```

For this and other results, please add that these should be non-negative values, and that any negative values arising due to noise should be clamped to zero.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 99 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqDvIK5NBTzmwmyfuu:-OPqDvIK5NBTzmwmyfuv:b3493v0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L99)):*
> ```proto
> 
>     // Interval for when data is guaranteed to be available for a
>     // [Requisition][] fulfillment.
> ```

Append: using the `ModelLine` named by `key`
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 101 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqEEwACd7dwQryC2Zm:-OPqEEwACd7dwQryC2Zn:bbek4wt) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L101)):*
> ```proto
>     // [Requisition][] fulfillment.
>     //
>     // This specifically means that if a [Requisition][] were to be created, the
> ```

This is subject to the requisition's interval too.

Similar to below, it might make sense to lift all of the discussion of requisition to the requisitions service, and try to focus on clearly defining the semantics of the resources in a more isolated way.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: 0 of 7 files reviewed, 7 unresolved discussions (waiting on @kungfucraig)

___
*[a discussion](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqCSuB1cPHjfGpYUeq:-OPqPrQTDOqyhbTDRdfx:b-cu91xq) (no related file):*

It would be useful to add some additional context to requisitions_service.proto referencing the resource documentation where necessary. The service definitions are a like entry point for people.
</blockquote></details>

I made a note of the protocols for each method. It's a resource-oriented API, so looking at the resource definition is pretty important. The only thing I can think to add is that the Requisition must be fulfilled according to the RequisitionSpec and MeasurementSpec, but that feels pretty self-explanatory/redundant.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 99 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqDvIK5NBTzmwmyfuu:-OPqWygzClijwIo1dZEl:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L99)):*

Append: using the `ModelLine` named by `key`
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 101 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqEEwACd7dwQryC2Zm:-OPqXtD--4XoUx8Xtg9g:b10cvoi) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L101)):*

This is subject to the requisition's interval too.

Similar to below, it might make sense to lift all of the discussion of requisition to the requisitions service, and try to focus on clearly defining the semantics of the resources in a more isolated way.
</blockquote></details>

I'm trying to capture what is effectively the VID labeling requirement, i.e. what an EDP is promising by specifying this interval. I tried to rephrase.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 171 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqDPvj26fs9c5eo_j9:-OPqXsT0-M6p1jDssGa_:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L171)):*

Adding the "subject to" doesn't provide clarity. This is the availability of events for the Event Group. How it relates to DP data availability is a function of requisition fulfillment.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 181 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPpt2jD4-a9jSgwAVsl:-OPqXOaf1CvBW1D3nIK7:bkeg7qr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L181)):*

s/If/When so that the documentation will age better when we make it required.
</blockquote></details>

Done, though really the conditional part should be deleted from the comment at the same time the field is marked as REQUIRED.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/measurement.proto` line 189 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPqCkfE7hBVQdiW19jm:-OPqvUD2BqrGsoDJad-E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/cdb6ea089b56cd0b43196e547cc456772cdcf2b2/src/main/proto/wfa/measurement/api/v2alpha/measurement.proto#L189)):*

For this and other results, please add that these should be non-negative values, and that any negative values arising due to noise should be clamped to zero.
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition.proto` line 256 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OPr2DqC3Cv1Rpt9ehzB:-OPr2DqC3Cv1Rpt9ehzC:bda5fm8) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/22bbad950887c08cbc7f4e32e16e81ae0acd716d/src/main/proto/wfa/measurement/api/v2alpha/requisition.proto#L256)):*
> ```proto
>     // Justification for refusing to fulfill this [Requisition][].
>     //
>     // Note that there is no [Justification][] for
> ```

@kungfucraig I figured this is the most explicit we can be to make it clear that refusing under this condition is incorrect.

kungfucraig (approved):
> Reviewed 6 of 7 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`-- commits` line 4 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdYzRH03LL4MKlvdA_:-OQdYzRH03LL4MKlvdAa:be3gyu2):*
shouldn't the reporting server prevent this from happening?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`-- commits` line 4 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdYzRH03LL4MKlvdA_:-OQdbcFG5Y9zXEnTh14-:b9et9ju):*

shouldn't the reporting server prevent this from happening?
</blockquote></details>

This is referring to incorrect EDP requisition fulfillment behavior, i.e. EDPs incorrectly interpreting the API and refusing requisitions they should fulfill.

stevenwarejones (changes requested):
> Reviewed 6 of 7 files at r2, 1 of 1 files at r3.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 4 unresolved discussions (waiting on @SanjayVas)

___
*[`-- commits` line 4 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdYzRH03LL4MKlvdA_:-OQdjhJI5qx3pNTNL-NN:bxagrvx):*

This is referring to incorrect EDP requisition fulfillment behavior, i.e. EDPs incorrectly interpreting the API and refusing requisitions they should fulfill.
</blockquote></details>

okay - misread this the first time.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQdkJqi6pPAFHusqkKl:bj3awin) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*
> ```proto
>   //
>   // If this [EventGroup][] represents an ad campaign, this would be considered
>   // the start and end times of the campaign.
> ```

I think there's a problem with this comment. The start and end of the campaign may not coincide with its availability within the local market platform.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdjz810TECN9FJBdED:-OQdjz810TECN9FJBdEE:bafpajr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L103)):*
> ```proto
>     // The guarantee is that for any [Requisition][] where
>     // [MeasurementSpec.model_line][] matches [key][], the [DataProvider][] has
>     // the necessary data to fulfill it when the interval specified by
> ```

I think the phrase that it is guaranteed that a data provider has the necessary data to fulfill this is confusing. What is trying to be communicated? Just that it should not be rejected?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdksC_2ooX3Bvq-GVW:-OQdksC_2ooX3Bvq-GVX:b9hn4sq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L37)):*
> ```proto
>   // See [Requisitions.FulfillDirectRequisition][] for fulfilling a
>   // [Requisition][] using the [Direct][ProtocolConfig.Direct] protocol.
>   rpc FulfillRequisition(stream FulfillRequisitionRequest)
> ```

it was previously supported to direct fulfill to a duchy. Are we deprecating that?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdjz810TECN9FJBdED:-OQeMjaF0r9aX-IzCaYQ:b-xl3c3f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L103)):*

I think the phrase that it is guaranteed that a data provider has the necessary data to fulfill this is confusing. What is trying to be communicated? Just that it should not be rejected?
</blockquote></details>

It's effectively trying to indicate that the EDP needs to label all events in all EventGroups in the stated ranges with the stated model lines. I'm trying to keep things in terms of the API though, and VID labeling is an implementation detail on the EDP side.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQeY4ku1BchGHyH8M1m:b-qkgu63) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

I think there's a problem with this comment. The start and end of the campaign may not coincide with its availability within the local market platform.
</blockquote></details>

I think the intent is for this to always match the campaign start/end, and to control the availability within the local market platform using the availability interval at the DataProvider level. The implication is that EDPs can't create EventGroups until they've had their event labeling backfilled. CC @kungfucraig

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdksC_2ooX3Bvq-GVW:-OQeYXJe7enDWczOLavn:b-d632yu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L37)):*
> it was previously supported to direct fulfill to a duchy

I'm not sure what you're referring to. The direct protocol by definition means fulfilling directly at the Kingdom, bypassing Duchies. The other protocols go to the Duchy.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdjz810TECN9FJBdED:-OQf3TsB9r63GLHAzQFX:b-ptislb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L103)):*

It's effectively trying to indicate that the EDP needs to label all events in all EventGroups in the stated ranges with the stated model lines. I'm trying to keep things in terms of the API though, and VID labeling is an implementation detail on the EDP side.
</blockquote></details>

maybe something along the lines of "Data Provides should only set availability for an interval they are guaranteed to have data to fulfill any requested requisition within that interval."
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQf3qNqBhVsuEQCzrXK:b-5ccctw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

I think the intent is for this to always match the campaign start/end, and to control the availability within the local market platform using the availability interval at the DataProvider level. The implication is that EDPs can't create EventGroups until they've had their event labeling backfilled. CC @kungfucraig
</blockquote></details>

Backfilling is problematic. I don't think future EDPs will be able to backfill. Or even if many can, that all will be able to.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdksC_2ooX3Bvq-GVW:-OQf44PAAeDpMBEtQ2m0:ba3to8s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L37)):*

> it was previously supported to direct fulfill to a duchy

I'm not sure what you're referring to. The direct protocol by definition means fulfilling directly at the Kingdom, bypassing Duchies. The other protocols go to the Duchy.
</blockquote></details>

There was a previous ability to fufill direct measurements directly to duchies. We may have already deprecated that ability. Consider this comment non-blocking as the comment is fine from my POV.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQfLx9t7pQG6dk0s4xu:bicuh3s) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

Backfilling is problematic. I don't think future EDPs will be able to backfill. Or even if many can, that all will be able to.
</blockquote></details>

I thought this was part of the "always on" requirement.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQiiP78EkD8AIBCiNxl:b-hcjo5p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

I thought this was part of the "always on" requirement.
</blockquote></details>

what if an advertiser onboards a campaign that is 2 year long. I don't think this requirement can be this open-ended.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQjZiMD-bZRIJofk1p2:b-vlp1jv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

what if an advertiser onboards a campaign that is 2 year long. I don't think this requirement can be this open-ended.
</blockquote></details>

Maybe we write "typically this would be ..."? I can definitely see where this may not always be the case, but for the vast majority of the cases this seems reasonable.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: 5 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 103 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdjz810TECN9FJBdED:-OQjoCsf3_UW3hzZx2YR:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L103)):*

maybe something along the lines of "Data Provides should only set availability for an interval they are guaranteed to have data to fulfill any requested requisition within that interval."
</blockquote></details>

Done.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQjnOotAIu3uJym37C0:b-pbs2d2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

Maybe we write "typically this would be ..."? I can definitely see where this may not always be the case, but for the vast majority of the cases this seems reasonable.
</blockquote></details>

Done after confirming with @kungfucraig that backfilling requirements are up to the local market.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 37 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdksC_2ooX3Bvq-GVW:-OQjoIuD8L1Hr-qR2iiN:b-vvtlcq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L37)):*
> There was a previous ability to fufill direct measurements directly to duchies

This has never been the case. You may be thinking of the brief period where EDPs were able to choose between either using the Direct protocol or the LLv2 protocol for single-EDP R/F Measurements.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQtOmj58y1PjXEzesDt:b-kgi2w9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

Done after confirming with @kungfucraig that backfilling requirements are up to the local market.
</blockquote></details>

I'm not even sure with the wording of typically. I'd like confirmation from multiple EDPs that this is feasible and scalable first.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQtTNBPDiMOcjPBo1Jg:b-5osavn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

I'm not even sure with the wording of typically. I'd like confirmation from multiple EDPs that this is feasible and scalable first.
</blockquote></details>

Note that we're *not* indicating that backfill is required here. The "typically" is to indicate that we expect these dates to be the campaign start and end date in most cases. The exception is cases like the one you mentioned where an EDP exports an EventGroup for an older campaign and does not backfill.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 102 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQtuCRJ5qIRlx6dwPjJ:-OQu0_O_CIcfxfVjU3AF:bm2wc3j) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2e6a0bd8184aff8e62358dac72e87074d3b577a0/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L102)):*

maybe `to fulfill any requistiion within that interval unless other specified in that Event Group's data availability intervals.`
</blockquote></details>

That's covered by the "full any Requisition" part, as the documentation for RequisitionSpec explains how the collection interval works w.r.t. the two data availability fields.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 182 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQtuOT52j85J5cAjFi4:-OQu-fcg-J6agyhZT3OL:bdpc2g6) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2e6a0bd8184aff8e62358dac72e87074d3b577a0/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L182)):*

remind me why we don't use `DataAvailabilityMapEntry` here as well?
</blockquote></details>

We don't want to support per-ModelLine availability at the EventGroup level. See #226.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 178 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk:-OQuIV2j5INQ-DynSawi:b-3ri6bv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ef50c91c0cd2827ff9a0d0bf0561e9501ca03c53/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L178)):*

Note that we're *not* indicating that backfill is required here. The "typically" is to indicate that we expect these dates to be the campaign start and end date in most cases. The exception is cases like the one you mentioned where an EDP exports an EventGroup for an older campaign and does not backfill.
</blockquote></details>

To be clear this PR is a documentation clean up not a refactor.

As for the design choice, we discussed this issue at some length in the past and I believe we concluded that 1) it didn't look like anyone needed this; 2) is complicates things quite a bit; 3) if it was ever needed that adding it in the future was not much harder than adding it right now would be. Thus it made sense to start with the simpler approach.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 182 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQtuOT52j85J5cAjFi4:-OQyFc3u-1N_kZbmUC8K:b-ckqba) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2e6a0bd8184aff8e62358dac72e87074d3b577a0/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L182)):*

We don't want to support per-ModelLine availability at the EventGroup level. See #226.
</blockquote></details>

i think we should probably revisit this - I don't see how Meta, for example, could implement this without per ModelLine availability

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 102 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQtuCRJ5qIRlx6dwPjJ:-OQyFkEm6D6Tp8-2emVx:b2ueczp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2e6a0bd8184aff8e62358dac72e87074d3b577a0/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L102)):*

That's covered by the "full any Requisition" part, as the documentation for RequisitionSpec explains how the collection interval works w.r.t. the two data availability fields.
</blockquote></details>

I'd prefer the documentation be explicit in both places

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: all files reviewed, 2 unresolved discussions (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 182 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQtuOT52j85J5cAjFi4:-OQyWXOJ92RjPQD9QYql:b2lfbzq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2e6a0bd8184aff8e62358dac72e87074d3b577a0/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L182)):*

i think we should probably revisit this - I don't see how Meta, for example, could implement this without per ModelLine availability
</blockquote></details>

Let's handle this discussion elsewhere, as this PR is just clarifying the design decision that was already made (and in fact already discussed with you in terms of Meta's ability to support it).

See Craig's comment https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQdkJqi6pPAFHusqkKk

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: 6 of 7 files reviewed, 2 unresolved discussions (waiting on @kungfucraig and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto` line 102 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237#-OQtuCRJ5qIRlx6dwPjJ:-OQyXoHLAOAXKIOdRzsV:b-74byqr) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/2e6a0bd8184aff8e62358dac72e87074d3b577a0/src/main/proto/wfa/measurement/api/v2alpha/data_provider.proto#L102)):*

I'd prefer the documentation be explicit in both places
</blockquote></details>

I added a note to see the collection interval field. Understanding requisition fulfillment is a prerequisite to setting this field.

stevenwarejones (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/237)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #242, opened 2025-06-16
feat: Add TrusTEE protocol config and requisition fulfillment.

renjiezh:


renjiezh:
https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/2453

SanjayVas (changes requested):
> Reviewed 3 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 4 unresolved discussions (waiting on @kungfucraig and @roaminggypsy)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 112 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTU9gYwC5hSuvBqGUAd:-OTU9gYwC5hSuvBqGUAe:b-9vtg1w) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e6542f4179380ef60c2e0991493b8a62ddff65b6/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L112)):*
nit: this is generally referred to as a workload identity provider, which is a part of the workload identity federation concept. I think HashiCorp has a clear general explanation[^1]:

> Configure a workload identity provider when you want to create a trust relationship between HashiCorp Cloud Platform (HCP) and an external identity provider. This trust relationship is called _federation_.

[^1]: https://developer.hashicorp.com/hcp/docs/hcp/iam/service-principal/workload-identity-federation



_Suggestion:_
```proto
        // Resource name of the workload identity provider.
        string workload_identity_provider = 3;
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 105 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTUC2-V1vpsPlGXO93A:-OTUC2-W7fmUja5egYDd:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e6542f4179380ef60c2e0991493b8a62ddff65b6/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L105)):*
> ```proto
>       // Specification used to access the KMS and decrypt the data.
>       message EncryptionSpec {
>         // An encrypted data encryption key (dek), which is in the format of
> ```

nit



_Suggestion:_
```proto
DEK
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 106 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTUC9RO2AWH5DnxZE3t:-OTUC9RO2AWH5DnxZE3u:bh2h59f) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e6542f4179380ef60c2e0991493b8a62ddff65b6/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L106)):*
> ```proto
>       message EncryptionSpec {
>         // An encrypted data encryption key (dek), which is in the format of
>         // AES-256-GCM.
> ```

Add a structure to src/main/proto/wfa/measurement/api/v2alpha/crypto.proto for this similar to `EncryptionPublicKey`, but for keys that may be symmetric. This way we can more easily swap wire formats in case we switch away from Tink.

I assume this is going to be a serialized Tink keyset, so I'd expect a structure like

```proto
message EncryptionKey {
  // Format of data for an encryption key.
  enum Format {
    // Default value used if the format is omitted.
    FORMAT_UNSPECIFIED = 0;

    // Tink Encrypted Keyset format.
    //
    // [data][] is a serialized `google.crypto.tink.EncryptedKeyset` message. Ciphertexts use Tink's wire format.
    //
    // See https://developers.google.com/tink/wire-format
    TINK_ENCRYPTED_KEYSET = 1;
  }
  // Format of data for this key.
  Format format = 1 [(google.api.field_behavior) = REQUIRED];

  // The format-specific key data.
  //
  // [format][] and [data][] together must specify all necessary information to
  // decrypt ciphertexts.
  bytes data = 2 [(google.api.field_behavior) = REQUIRED];
}
```



_Code quote:_
```proto
        // An encrypted data encryption key (dek), which is in the format of
        // AES-256-GCM.
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 108 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTUC0Eo6jfVBtHQzxtX:-OTUC0Eo6jfVBtHQzxtY:b2ni8l0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/e6542f4179380ef60c2e0991493b8a62ddff65b6/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L108)):*
> ```proto
>         // AES-256-GCM.
>         bytes encrypted_data_encryption_key = 1;
>         // The Uri of the key encryption key (kek) in KMS that is used to
> ```

nit



_Suggestion:_
```proto
KEK
```

SanjayVas (approved):
> Reviewed 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig and @roaminggypsy)

kungfucraig (changes requested):
> Reviewed 2 of 3 files at r1, 2 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 2 unresolved discussions (waiting on @roaminggypsy)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTYIb9s4txyGjL1d42A:-OTYIb9s4txyGjL1d42B:b-ax8icy) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/fbdf8401033479b6718681c4d7b84d5f02496bb2/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L105)):*
> ```proto
>       // Specification used to access the KMS and decrypt the data.
>       message EncryptionSpec {
>         // An [EncryptionKey] in decrypted format used as data encryption key
> ```

Do you mean "encrypted format" instead of "decrypted format?"
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 110 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTYIf3nA4o0rx9TFa_d:-OTYIf3nA4o0rx9TFa_e:b-4ieavf) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/fbdf8401033479b6718681c4d7b84d5f02496bb2/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L110)):*
> ```proto
>         // It is Using AES-256-GCM algorithm for symmetric
>         // encryption/decryption.
>         EncryptionKey encrypted_data_encryption_key = 1;
> ```

To be consistent with the use of "kek" below, call this "encrypted_dek"

roaminggypsy (approved):
> Reviewed 2 of 3 files at r1, 1 of 2 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @roaminggypsy)

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 134 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OThbIVcBsHtZMgLNx_O:-OThbIVcBsHtZMgLNx_P:b5mhf26) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/76b3be52dafabaebbabb93ba76e14c33ecd21153/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L134)):*
> ```proto
>       DataFormat data_format = 2;
> 
>       // The SHA256 hash of the PopulationSpec.
> ```

Per conversation offline, let's make this a farm fingerprint of the PopulationSpec, which will be based on some deterministic stringification of it, but not the serialization.

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 2 unresolved discussions (waiting on @renjiezh and @roaminggypsy)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OThRHV9EqjWpC07AlHy:-OThwt6-7oR7OGew1LhE:b-13hetv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/76b3be52dafabaebbabb93ba76e14c33ecd21153/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*
<details><summary><i>Previously, roaminggypsy (Lin) wrote…</i></summary><blockquote>

One question: why is unspecified the default value instead of frequency vector?
</blockquote></details>

We almost always use UNSPECIFIED as the 0/default value. Encoding a true default has been problematic as it makes changing it a backwards-incompatible change (where it's also unclear where the default might be applied, on the client vs. the server). Back in proto2, protobuf itself used to allow you to specify default values for any field. This was dropped in proto3.

roaminggypsy (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 1 unresolved discussion (waiting on @renjiezh)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @roaminggypsy, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/crypto.proto` line 105 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTi0xKCEF7AFg30M8Lh:-OTi1tgV4LfHOh8i_zby:b54f4mw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/76b3be52dafabaebbabb93ba76e14c33ecd21153/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto#L105)):*
<details><summary><i>Previously, roaminggypsy (Lin) wrote…</i></summary><blockquote>

Nit: The comment seems incomplete
</blockquote></details>

Nice catch. Thanks
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 134 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OThbIVcBsHtZMgLNx_O:-OTi1v9v05TF7QgS-a9E:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/76b3be52dafabaebbabb93ba76e14c33ecd21153/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L134)):*

Per conversation offline, let's make this a farm fingerprint of the PopulationSpec, which will be based on some deterministic stringification of it, but not the serialization.
</blockquote></details>

Done.

kungfucraig (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @roaminggypsy and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OTm9eSED-E-r7aaTZ0l:b60x7pc) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*
> ```proto
> 
>         // Frequency vector in bytes.
>         FREQUENCY_VECTOR = 1;
> ```

Let's change this value to "ENCRYPTED_FREQUENCY_VECTOR" and add "UNENCRYPTED_FREQUENCY_VECTOR".

That will make it explicit when the payload should be treated as unencrypted vs. using the fact the encryption spec is not set.

@stevenwarejones does that address your comment in the design doc?

SanjayVas (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @roaminggypsy, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OTmTxDlEwN4SzhuduJM:bjvxf1q) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*

Let's change this value to "ENCRYPTED_FREQUENCY_VECTOR" and add "UNENCRYPTED_FREQUENCY_VECTOR".

That will make it explicit when the payload should be treated as unencrypted vs. using the fact the encryption spec is not set.

@stevenwarejones does that address your comment in the design doc?
</blockquote></details>

Minor not on that is I'd prefer `PLAINTEXT` over `UNENCRYPTED` if we need a prefix for it. Otherwise `FREQUENCY_VECTOR` and `ENCRYPTED_FREQUENCY_VECTOR` is also fine with me.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @kungfucraig, @roaminggypsy, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OTmpB6TDfAT2X_7TwQv:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*

Minor not on that is I'd prefer `PLAINTEXT` over `UNENCRYPTED` if we need a prefix for it. Otherwise `FREQUENCY_VECTOR` and `ENCRYPTED_FREQUENCY_VECTOR` is also fine with me.
</blockquote></details>

Done.

kungfucraig (changes requested):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 2 unresolved discussions (waiting on @roaminggypsy and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OTn6GQO42zbHjQvXVVh:b-jadneb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*
<details><summary><i>Previously, renjiezh wrote…</i></summary><blockquote>

Done.
</blockquote></details>

That works for me.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 138 at r6](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTn6WOh5NdC6c3S5z3d:-OTn6WOh5NdC6c3S5z3e:bxr5jkx) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/788b77397dd4c1df0178243d1d640930a9df47a8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L138)):*
> ```proto
>       DataFormat data_format = 2;
> 
>       // The fingerprint of the PopulationSpec.
> ```

Can you write the: "The farmhash fingerprint of the PopulationSpec"

TODO(...): Add a description of how to hash the PopulationSpec

stevenwarejones (changes requested):
> Reviewed 2 of 3 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @renjiezh)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OUfeXil85R4OsNGkMO3:bpf62e4) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*

That works for me.
</blockquote></details>

i would prefer encryption be covered in a separate field. There is DataFormat. And then there is EncryptionScheme which could be none.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 116 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUff5Lk-aN7rhmA9MMK:-OUff5Lk-aN7rhmA9MML:bn6wki5) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/96de7c47bb6f613d212ac51ffe9e09922df17253/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L116)):*
> ```proto
>         // Resource name of the workload identity provider to
>         // impersonate the service account with the KEK access.
>         string workload_identity_provider = 3;
> ```

is this needed @marco-premier ?

I thought stephen robinson had said this kind of impersonation is not needed also
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 119 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUff5pN3Q94cXdKqyyC:-OUff5pN3Q94cXdKqyyD:b-7ropln) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/96de7c47bb6f613d212ac51ffe9e09922df17253/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L119)):*
> ```proto
>         // The service account to impersonate to get an access token for the
>         // KEK.
>         string impersonated_service_account = 4;
> ```

ditto
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 123 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUfepB494owSOqUa8l6:-OUfepB494owSOqUa8l7:b-3pejsn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/96de7c47bb6f613d212ac51ffe9e09922df17253/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L123)):*
> ```proto
>       // The specification used to access the KMS and decrypt the data.
>       //
>       // If not specified, it implies the data is in plaintext.
> ```

i don't really prefer impled values. I would prefer a oneof or an enum if encryption was applied.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier and @renjiezh)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OUfhrqrBq6Dm6ksA5eq:b-bpgf0g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*

i would prefer encryption be covered in a separate field. There is DataFormat. And then there is EncryptionScheme which could be none.
</blockquote></details>

Then we're really just back to an optional EncryptionSpec or wrapping a one-of around it and have a sibling that is an empty message with the name "no_encryption."

I'd say that's a fine way to do it, but kind of a wash with respect to what is here.

Do you have something else in mind?

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig and @Marco-Premier)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 129 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OTm9eSED-E-r7aaTZ0k:-OUfq9RD7xzB8BXOMprU:b6ljivb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/f495c8546da977762ab25c18013e50a0fd4694d8/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L129)):*

Then we're really just back to an optional EncryptionSpec or wrapping a one-of around it and have a sibling that is an empty message with the name "no_encryption."

I'd say that's a fine way to do it, but kind of a wash with respect to what is here.

Do you have something else in mind?
</blockquote></details>

Thank you Steven for the idea. I updated the comments to make it clearer.

I feel DataFormat is a complete description of body chunks. By itself, people should know the exact shape of it, e.g, is it compressed, what compression algorithm is used., is it encrypted, what is the encryption algorithm. 

The EncryptionSpec is the additional and optional information if the data is encrypted. Regarding this field, it is either encrypted or not encrypted. There is no third option. So I think it is fine to make it optional and has implicit meaning.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 116 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUff5Lk-aN7rhmA9MMK:-OUftbvb8Cqi3P0Lfka-:bqvm2t0) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/96de7c47bb6f613d212ac51ffe9e09922df17253/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L116)):*

is this needed @marco-premier ?

I thought stephen robinson had said this kind of impersonation is not needed also
</blockquote></details>

According to my demo and official doc, these fields are required to do attestation at EDPs' Workload Identity Federation.

We assume EDPs manage their own WIF and KMS so we needs the information on the fly.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 3 of 4 files reviewed, 4 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, @renjiezh, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 123 at r8](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUfepB494owSOqUa8l6:-OUkEATl9KBj_uxrobz5:bk6ssvq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/96de7c47bb6f613d212ac51ffe9e09922df17253/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L123)):*

i don't really prefer impled values. I would prefer a oneof or an enum if encryption was applied.
</blockquote></details>

There's not actually an implication anymore, and the documentation should be revised. See my other comment.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 142 at r9](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUkEIjuARyPCEhrisvT:-OUkEIjuARyPCEhrisvU:b3stw2n) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/06851694f0bed344997a1761add98b5317741761/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L142)):*
> ```proto
>       // data is encrypted.
>       //
>       // If not specified, the data is in plaintext.
> ```

There's no implication here. Whether the data is encrypted is based solely on the DataFormat.



_Suggestion:_
```proto
// Required if [data_format][] is [DataFormat.ENCRYPTED_FREQUENCY_VECTOR][].
```

SanjayVas (approved):
> Reviewed 1 of 2 files at r5, 1 of 1 files at r10, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, @renjiezh, and @stevenwarejones)

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 3 of 4 files reviewed, 1 unresolved discussion (waiting on @kungfucraig, @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/crypto.proto` line 119 at r10](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OUltHtAANljdJ9IwSKI:-OV9Ohxo8x_AH_PsZQ4a:b-z97y2r) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/638b3048823cf8d77339bc2a5d4f3ba213f61bd3/src/main/proto/wfa/measurement/api/v2alpha/crypto.proto#L119)):*

I'm not sure why this would be an issue. We already use Tink keysets in this API.
</blockquote></details>

KeyDerivation added.

SanjayVas (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: 3 of 4 files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 105 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OV9vTl6BfgAnBTrjefp:-OV9vTl6BfgAnBTrjefq:b-fttthu) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ff0e9b9e13525eda6afcc9f2f39133970ef90385/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L105)):*
> ```proto
>       // Specification for envelope decryption, where the Data Encryption Key
>       // (DEK) is sent in an encrypted form.
>       message EnvelopeDecryption {
> ```

The general concept is referred to as envelope *encryption* even if the Duchy is only doing the decryption part.



_Suggestion:_
```proto
EnvelopeEncryption
```
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 121 at r11](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OV9vos-AON62jQRLYKY:-OV9vos027B4wfauGACz:b-ihumzj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/ff0e9b9e13525eda6afcc9f2f39133970ef90385/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L121)):*
> ```proto
> 
>       // Specification for key derivation, where the Data Encryption Key
>       // (DEK) is derived from a master key.
> ```

I don't think this is the correct terminology here. I have to look more into the implementation, but I assume what we're specifically wanting to implement is HMAC using some KMS-managed aspect.



_Code quote:_
```proto
from a master key.
```

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OVJwR2iA0emJQaX543N:-OVJwuOw4McDPiXU2CN4:b-byesaz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/538d96fcdca6b4ab94ac3b75a6325875a2557cf5/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L153)):*

why do we need this fingerprint?
</blockquote></details>

Good question. 
This is a new feature we wish we had at beginning of HM shuffle. It is to make sure all EDPs are using the same population_spec. A function in the EDP library will be provided to calculate the fingerprint.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @renjiezh, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OVJwR2iA0emJQaX543N:-OVKPTP33LvoDpl-_wT6:bbdhk2d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/538d96fcdca6b4ab94ac3b75a6325875a2557cf5/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L153)):*
<details><summary><i>Previously, renjiezh wrote…</i></summary><blockquote>

Good question. 
This is a new feature we wish we had at beginning of HM shuffle. It is to make sure all EDPs are using the same population_spec. A function in the EDP library will be provided to calculate the fingerprint.
</blockquote></details>

A couple issues
1. If its part of the requisition spec, i don't see why every party needs to verify it. What if a party uses the wrong vid model? Should we be including that hash as well?
2. the population spec is part of the frequency vector. If we were to use a different sketch format in the future, this field won't work. I'd prefer a schema where this is only used for some data formats.

renjiezh (commented):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @kungfucraig, @Marco-Premier, and @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OVJwR2iA0emJQaX543N:-OVKWU7X2pYPyaVQgbzO:bnjj2qo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/538d96fcdca6b4ab94ac3b75a6325875a2557cf5/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L153)):*

A couple issues
1. If its part of the requisition spec, i don't see why every party needs to verify it. What if a party uses the wrong vid model? Should we be including that hash as well?
2. the population spec is part of the frequency vector. If we were to use a different sketch format in the future, this field won't work. I'd prefer a schema where this is only used for some data formats.
</blockquote></details>

Hi @kungfucraig, what do you think about this field?

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OVJwR2iA0emJQaX543N:-OVNvxeYCFU7p29LiO4C:b-og88sv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/538d96fcdca6b4ab94ac3b75a6325875a2557cf5/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L153)):*
<details><summary><i>Previously, renjiezh wrote…</i></summary><blockquote>

Hi @kungfucraig, what do you think about this field?
</blockquote></details>

In Origin tests of HMSS we have seen mismatched population specs as a frequent source of problems. They are uncovered by another error, "mismatched frequency vector size." I'd like to be able to detect this more explicitly so the system can provide better guidance for remediation.

As to a hash of the VID model, there really is no common thing that each EDP could hash.

The population spec is not part of the frequency vector, but rather the size of frequency vector and the VID mapping is derived from the population spec.

Once we have full support for the model repo in place, where the EDP gets the population spec from the ModelRelease linkage, it's possible we will not need this. But even then there's no harm.

kungfucraig (approved):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 3 unresolved discussions (waiting on @Marco-Premier, @SanjayVas, and @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 153 at r12](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242#-OVJwR2iA0emJQaX543N:-OVUK4UO2R_3yoD9M51G:b-9oovqp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/538d96fcdca6b4ab94ac3b75a6325875a2557cf5/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L153)):*

I'm also wondering if this hash is only relevant to frequency_vectors then either this comment needs to be updated or the schema updated so that its obvious to only populate.... for example, you could add a message of FrequencyVectorDetails that contains this. and then there could be a oneof of FrequencyVEctorDetails vs details for other sketch formats
</blockquote></details>

At this time there are no other supported sketch formats. I think we can safely revisit this if/when we cross the bridge.

stevenwarejones (approved):
> Reviewed 1 of 1 files at r12.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: all files reviewed, 2 unresolved discussions (waiting on @Marco-Premier and @SanjayVas)

SanjayVas (approved):
> Reviewed 1 of 1 files at r12, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/242)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @renjiezh)

PR #247, opened 2025-08-05
feat: Add version fields for EventFieldDescriptor and a new EventDescriptor with a version field.

kungfucraig:


SanjayVas (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/247)* status: all files reviewed, 1 unresolved discussion (waiting on @stevenwarejones)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_annotations.proto` line 96 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/247#-OWwZDuy1WYgbOuzP2Vz:-OWwZDuy1WYgbOuzP2W-:b-38rlkn) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6e2e9040ad267611f034321e659af6986f9cd533/src/main/proto/wfa/measurement/api/v2alpha/event_annotations.proto#L96)):*
> ```proto
> // Descriptor for an Event message type.
> message EventDescriptor {
>   // The current version of the Event. Must be positive.
> ```

nit: Maybe also note that updates to this value must be monotonically increasing for the same event message?
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_annotations.proto` line 105 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/247#-OWwY_dI1dHMvaQDhpgV:-OWwY_dI1dHMvaQDhpgW:beios7c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6e2e9040ad267611f034321e659af6986f9cd533/src/main/proto/wfa/measurement/api/v2alpha/event_annotations.proto#L105)):*
> ```proto
>   EventTemplateDescriptor event_template = 50000;
>   EventFieldDescriptor event_field = 50001 [deprecated = true];
>   EventDescriptor event = 50002;
> ```

FYI this is the first time that the Event message itself must be explicitly defined for a set of templates. So far we've defined one for convenience, but it's not been strictly required to be the same message. In theory, each EDP could have define their own event message as long as it had fields for all of the template messages with the right names.

stevenwarejones (approved):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/247)* status: all files reviewed, 1 unresolved discussion (waiting on @kungfucraig)

kungfucraig (commented):
> Reviewed 1 of 1 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/247)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @kungfucraig)

PR #248, opened 2025-08-06
feat: Annotate fields with future disposition

SanjayVas:
Issue: world-federation-of-advertisers/cross-media-measurement#2728

### DIFF
--- a/src/main/proto/api-linter.yaml
+++ b/src/main/proto/api-linter.yaml
@@ -14,6 +14,7 @@
 
 - included_paths:
   - 'wfa/measurement/api/v2alpha/**/*.proto'
+  - 'wfa/measurement/type/**/*.proto'
   disabled_rules:
   - 'core::0191::java-package'
   - 'core::0127::http-annotation'
--- a/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
+++ b/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
@@ -108,6 +108,7 @@ proto_library(
     deps = [
         ":crypto_proto",
         ":measurement_consumer_proto",
+        "//src/main/proto/wfa/measurement/type:future_disposition_proto",
         "@com_google_googleapis//google/api:field_behavior_proto",
         "@com_google_googleapis//google/api:resource_proto",
         "@com_google_googleapis//google/type:interval_proto",
@@ -133,6 +134,7 @@ proto_library(
         ":event_group_metadata_proto",
         ":event_template_proto",
         ":media_type_proto",
+        "//src/main/proto/wfa/measurement/type:future_disposition_proto",
         "@com_google_googleapis//google/api:field_behavior_proto",
         "@com_google_googleapis//google/api:resource_proto",
         "@com_google_googleapis//google/type:interval_proto",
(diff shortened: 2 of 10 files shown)


stevenwarejones (changes requested):
> @stevenwarejones reviewed 10 of 10 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/248)* status: all files reviewed, 2 unresolved discussions (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 153 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/248#-OYH9Qe06bqloX-l__4W:-OYH9Qe06bqloX-l__4X:b-f0atxa) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9c7824973f3329c48c22bdf29e8bea44d870f468/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L153)):*
> ```proto
>   // used to last update this resource.
>   //
>   // This field will be deprcated in a future release. See `encrypted_metadata`
> ```

nit: spelling on `deprecated`
___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisitions_service.proto` line 165 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/248#-OYHAUqU4Qrs1Kzpm2bh:-OYHAUqU4Qrs1Kzpm2bi:bkm8zh9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9c7824973f3329c48c22bdf29e8bea44d870f468/src/main/proto/wfa/measurement/api/v2alpha/requisitions_service.proto#L165)):*
> ```proto
>   //
>   // This may be required in the future.
>   EncryptedMessage encrypted_result = 4 [(type.future_disposition) = REQUIRED];
> ```

this is an awkward one to tag as required in the future since we also already intend to fully deprecate it in the future.

stevenwarejones (changes requested):
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/248)* status: 9 of 10 files reviewed, 1 unresolved discussion (waiting on @SanjayVas)

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisitions_service.proto` line 165 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/248#-OYHAUqU4Qrs1Kzpm2bh:-OYWNOdP4yNUjZ6EBEAY:bsejybj) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/9c7824973f3329c48c22bdf29e8bea44d870f468/src/main/proto/wfa/measurement/api/v2alpha/requisitions_service.proto#L165)):*

Yes, that's the unfortunate situation we're in by deprioritizing getting new releases out for finalizing API changes. We're effectively 2 minor releases behind where we should be.

Note that this field wouldn't be removed until v0.8.0 with dropping consent signaling, so this is still valid for some time yet.
</blockquote></details>

also - the language on the comment is inappropriate imo. `this may be required in the future` vs the disposition that it _will_ be required

PR #249, opened 2025-08-26
feat: Add missing methods and fields for Populations

SanjayVas:
* Add UpdatePopulation for updating blob URIs.
* Add LookupPopulation for looking up Population resources by blob URI.
* Deprecate Population.event_template.
* Add Population.PopulationBlob.blob_type_url to handle version differences.

Issue: world-federation-of-advertisers/cross-media-measurement#2835
Issue: world-federation-of-advertisers/cross-media-measurement#2837


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/249)

PR #251, opened 2025-09-15
feat: Remove restriction on retroactive ModelLines and ModelRollouts

SanjayVas:
Issue: world-federation-of-advertisers/cross-media-measurement#2920

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/model_line.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/model_line.proto
@@ -44,8 +44,7 @@ message ModelLine {
   string description = 3;
 
   // The start of the time range when this `ModelLine` is active (available for
-  // labeling), inclusive. When specified in a request message, this must be in
-  // the future.
+  // labeling), inclusive.
   google.protobuf.Timestamp active_start_time = 4 [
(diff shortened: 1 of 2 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/251)

SanjayVas (commented):
> This does, of course, only work if EDPs in question are willing to backfill

As noted in the corresponding issue, this API change shouldn't require EDPs to backfill since they can specify their data availability per-ModelLine. This means that an EDP can have data availability which starts after the rollout period.

> Can't we just make a one-time exception for the launch of the service and not make this comment change?

This is not just for initial launch. It also applies going forward for cases such as the K8s test, which will need to be able to create new ModelRollouts for new ModelLines when the PopulationSpec changes. The same may be true depending on what future rounds of synthetic data testing are desired.

> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/251#-:-O_IgiTQ6lxquMBn2J0x:b-9d5bt9)* status: 1 of 2 files reviewed, all discussions resolved (waiting on @stevenwarejones)

PR #254, opened 2025-10-01
docs: Clarify which intervals EnumerateValidModelLines considers

SanjayVas:
The primary use case of the method is for a MeasurementConsumer caller to determine which ModelLine to specify for a Measurement. For this purpose, only the DataProvider data availability for each ModelLine is relevant. The active range of the ModelLine is only transitively considered by the fact that DataProvider data availability is inherently constrained by the ModelLine active range.

Issue: world-federation-of-advertisers/cross-media-measurement#2998


kungfucraig (approved):
> @kungfucraig reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/254)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @SanjayVas)

PR #256, opened 2025-10-01
feat!: Add wildcard support and active interval filter to ListModelLines

SanjayVas:
Issue: world-federation-of-advertisers/cross-media-measurement#2997

BREAKING-CHANGE: The `ListModelLinesRequest.Filter.type` field has been renamed to `type_in` to be consistent with newer structured filter messages. This is a binary-compatible change.


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/256)

PR #274, opened 2026-04-10
feat: Add support for EventGroup hierarchies

SanjayVas:
The new EventGroupChain pseudo-resource provides a denormalized view of a leaf EventGroup and its parents. This can make integrations more convenient, for example creating a whole chain of EventGroups at once.

## Corresponding methods (TODO)

* LookupEventGroup - Looks up an EventGroup by its entity path
  * Could alternatively just add an entity path filter to ListEventGroups.
* CreateEventGroupChain - Creates a new chain, specifically a new leaf.
  * Any non-existent ancestors will also be created.
* UpdateEventGroupChain - Updates all EventGroups in a chain.
  * Any non-existent ancestors will be created. This cannot be used to add a new leaf.
* LookupEventGroupChain - Looks up a chain by its entity path (entity keys from root to leaf).

The EventGroupChain mutators could rely on the entity key rather than the name, so you need not know if an ancestor EventGroup already exists. This means it can fill in the `parent_event_group` fields for you based on the order of the specified EventGroups.

## Caveats

* The EventGroupChain design precludes adding support for multiple parents, since it relies on uniquely identifying a chain by the leaf EventGroup.

## Phased rollout

* The default value of the `entity_type_in` filter field for ListEventGroups ensures backwards-compatibility for existing clients.
* A given Halo instance (Kingdom) would have an option (flag) indicating what entity types it supports. This defaults to CAMPAIGN for backwards compatibility with existing EventGroups.
  * It is an error to create an EventGroup with an unsupported type.
  * Due to the restriction requiring each EventGroup in a chain to have a unique entity type, this inherently means that hierarchical EventGroups are not supported until a Kingdom supports more than one entity type.
  * The operator should only update this once its UI supports a given entity type. It is an error to remove an entity type from the option once an EventGroup has been created with that type.

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
+++ b/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
@@ -159,7 +159,9 @@ proto_library(
     srcs = ["event_group_metadata.proto"],
     strip_import_prefix = IMPORT_PREFIX,
     deps = [
+        "//src/main/proto/wfa/measurement/type:future_disposition_proto",
         "@com_google_googleapis//google/api:field_behavior_proto",
+        "@com_google_protobuf//:struct_proto",
     ],
 )
 
@@ -180,6 +182,17 @@ proto_library(
         "@com_google_googleapis//google/api:field_behavior_proto",
         "@com_google_googleapis//google/api:resource_proto",
         "@com_google_protobuf//:descriptor_proto",
+    ],
+)
+
+proto_library(
+    name = "event_group_chain_proto",
+    srcs = ["event_group_chain.proto"],
+    strip_import_prefix = IMPORT_PREFIX,
+    deps = [
+        ":event_group_proto",
+        "@com_google_googleapis//google/api:field_behavior_proto",
+        "@com_google_googleapis//google/api:resource_proto",
     ],
 )
 
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
@@ -60,8 +60,9 @@ message EventGroup {
 
   // ID referencing the `EventGroup` in an external system, provided by the
   // `DataProvider`. This value should be used to help the `DataProvider`
-  // synchronize their metadata with the EventGroupsService. See
-  // `EventGroupMetadata` which provides a set of identifiers with better
+  // synchronize their metadata with the EventGroupsService.
+  //
+  // See [entity_key][] which provides a set of identifiers with better
   // defined semantics.
   string event_group_reference_id = 5;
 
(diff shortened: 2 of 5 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274)

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274)* status: 0 of 4 files reviewed, 1 unresolved discussion.

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_chain.proto` line 57 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-OptgUzV1E0BbjhK5quy:-OptgUzV1E0BbjhK5quz:b-gp51cm) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/6077734b44066c50a2a83028bf967f20f1402c41/src/main/proto/wfa/measurement/api/v2alpha/event_group_chain.proto#L57)):*
> ```proto
>   ];
> 
>   // Metadata for each [EventGroup][] making up this chain from root to leaf.
> ```

Alternatively if we enforce that each key must be unique across all entity types for a given EDP, this can be a flattened merged structure.

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274)* status: 0 of 4 files reviewed, 2 unresolved discussions.

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 229 at r2](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-OptpC3_3ylqQaWhubcQ:-OptpC3_3ylqQaWhubcR:b-ro743p) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/982075342362746e9b49ce1e4ebba483ba9545b6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L229)):*
> ```proto
>     // DV360 line item.
>     LINE_ITEM = 2;
>     // Google Ads Ad Group.
> ```

This is just the initial meaning of this type. It can be expanded if e.g. another EDP wants to use the same entity type.

SanjayVas (commented):
> @SanjayVas resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274)* status: 0 of 5 files reviewed, 1 unresolved discussion.

SanjayVas:
See #275 for a non-hierarchical proposal.

stevenwarejones (commented):
it feels like this PR has 3 different ways of specifying relationships: parent_event_group, entity_path, and EventGroupChain. What is the reason for that?

> @stevenwarejones reviewed 2 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-:-OqN7oP19AEHT1dgGljO:b8tz00y)* status: 2 of 5 files reviewed, 1 unresolved discussion.

SanjayVas (commented):
`parent_event_group` is the canonical one that drives everything else. `entity_path` is OUTPUT_ONLY and is derived from `parent_event_group`.

`EventGroupChain` is a pseudo-resource with two purposes:

1. Get a denormalized view of an entire chain represented by a leaf EventGroup
2. Be able to manage chains of EventGroups knowing only their relationships (i.e. make a single call to create/update an entire chain)

See the PR description.

#275 takes a different approach prioritizing time-to-market of granular EventGroups over representing hierarchical EventGroups in the API up-front. That approach involves denormalizing the entire hierarchy into each leaf EventGroups, not bothering to maintain EventGroups for ancestors. The `event_type_in` filter is used to choose whether you're getting the existing campaign EGs or newer granular ones that represent the whole hiearchy. It's not my preferred solution, but the one that stakeholders want to go with for time-to-market.

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-:-OqNccBG18JoHBfRcN3E:b-daj3ts)* status: 2 of 5 files reviewed, 1 unresolved discussion (waiting on stevenwarejones).

stevenwarejones (commented):
Stepping back on the overall shape: I don't think EventGroupChain needs to be a resource. It has no independent lifecycle, its identity is just the leaf EventGroup's, and nothing references it — that's a denormalized projection, not a resource (which is what the embedded-resource linter suppression already says).

  A simpler approach that covers the same use cases:

  - Keep parent_event_group as a field on EventGroup, and make it mutable (not IMMUTABLE) — taxonomies evolve (inserting a new root or middle level re-parents existing nodes), so position has to be changeable.
  - Keep entity_path OUTPUT_ONLY, derived as parent's entity_path + this entity_key. Because it's derived, re-parenting one node cascades entity_path to the whole subtree server-side.
  - Add field-mask (partial update) support so a re-parent sends only parent_event_group. A single BatchUpdateEventGroups is already atomic, so bulk re-parent/re-order becomes one atomic, minimal-payload transaction — no read-modify-write, no clobbering the crypto fields. (Partial update is currently disabled because encrypted_metadata / measurement_consumer_public_key must be written together; scope the mask to structural fields so that invariant is
  preserved.)
  - Expose the denormalized chain view as a plain message returned by a convenience ListEventGroupChains RPC on EventGroupsService — no google.api.resource annotation, no name IDENTIFIER.

  This lets us drop CreateEventGroupChain, UpdateEventGroupChain(s), and LookupEventGroupChain entirely: mutation stays on EventGroup via Create/BatchCreate + field-masked BatchUpdate (a generic win, not hierarchy-specific), and the single-chain lookup-by-entity-path is just ListEventGroupChains with an entity_path filter. Middle-insert is then BatchCreate(new node) + one atomic field-masked BatchUpdate to re-parent the children; the only intermediate is a valid empty branch, never a broken/orphaned state.

> @stevenwarejones reviewed 1 file and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-:-OxgdfzUC1O_5SHG_Hzi:b7q0fn4)* status: 3 of 5 files reviewed, 2 unresolved discussions (waiting on SanjayVas).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 258 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-OxgYXbPCrzLA5h08-9U:-OxgYXbPCrzLA5h08-9V:baictdo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/648735f7231cead1390292d8fa2945723f740ee5/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L258)):*
> ```proto
>   // (-- api-linter: core::0121::no-mutable-cycles=disabled
>   //     aip.dev/not-precedent: Cycles cannot be formed. --)
>   string parent_event_group = 17
> ```

i think this field should be mutable so that EDPs can put in new taxonomy levels in the future.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto` line 249 at r5](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/274#-OxgY1sN5ylj8hNZrbYj:-OxgY1sN5ylj8hNZrbYk:b7gcdjo) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/648735f7231cead1390292d8fa2945723f740ee5/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto#L249)):*
> ```proto
>     // If unspecified, the server will treat this as if it contains
>     // [EventGroup.EntityType.CAMPAIGN][].
>     repeated EventGroup.EntityType entity_type_in = 11;
> ```

i assume you'll just use what's in main 
```
repeated string entity_type_in = 11;
```

PR #275, opened 2026-04-13
feat: Support more granular EventGroups

SanjayVas:
## Phased rollout

* All existing `EventGroup`s will have an `entity_type` of "campaign" for backwards compatibility.
  * This corresponds to the default value of the `entity_type_in` filter field in `ListEventGroups`.
* New granular `EventGroup`s will be created with an `entity_type` other than "campaign" and have `EventGroupMetadata.entity_metadata` specified.

## Potential future features

* Filters for `EventGroupMetadata.entity_metadata` in `ListEventGroups`.
* Representing the relationship between `EventGroup`s at different levels of granularity as a hierarchy.

Issue: world-federation-of-advertisers/cross-media-measurement#3697

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
+++ b/src/main/proto/wfa/measurement/api/v2alpha/BUILD.bazel
@@ -159,7 +159,9 @@ proto_library(
     srcs = ["event_group_metadata.proto"],
     strip_import_prefix = IMPORT_PREFIX,
     deps = [
+        "//src/main/proto/wfa/measurement/type:future_disposition_proto",
         "@com_google_googleapis//google/api:field_behavior_proto",
+        "@com_google_protobuf//:struct_proto",
     ],
 )
 
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto
@@ -60,8 +60,9 @@ message EventGroup {
 
   // ID referencing the `EventGroup` in an external system, provided by the
   // `DataProvider`. This value should be used to help the `DataProvider`
-  // synchronize their metadata with the EventGroupsService. See
-  // `EventGroupMetadata` which provides a set of identifiers with better
+  // synchronize their metadata with the EventGroupsService.
+  //
+  // See [entity_key][] which provides a set of identifiers with better
   // defined semantics.
   string event_group_reference_id = 5;
 
@@ -215,4 +216,30 @@ message EventGroup {
   // Only set when the view is `WITH_ACTIVITY_SUMMARY`.
   repeated AggregatedActivity aggregated_activities = 15
       [(google.api.field_behavior) = OUTPUT_ONLY];
+
+  // Key of the entity in the [DataProvider][]'s system that an [EventGroup][]
+  // represents.
+  message EntityKey {
+    // Type of the entity in the [DataProvider][]'s system as a human- and
+    // machine-readable string.
+    //
+    // This must be a URL-safe string, such that URL-encoding or -decoding
+    // results in the same string.
+    string entity_type = 1 [(google.api.field_behavior) = REQUIRED];
+
+    // ID of the entity in the [DataProvider][]'s system.
+    //
+    // This should be an ID that is available to [MeasurementConsumer][]'s
+    // through the [DataProvider][]'s system.
+    //
+    // This must be a URL-safe string, such that URL-encoding or -decoding
+    // results in the same string.
+    string entity_id = 2 [(type.future_disposition) = REQUIRED];
+  }
+  // Entity in the [DataProvider][]'s system that this [EventGroup][]
+  // represents.
+  //
+  // If specified, this must be unique across all [EventGroup][]s for the
+  // parent ([DataProvider][], [MeasurementConsumer][]) pair.
+  EntityKey entity_key = 16 [(type.future_disposition) = REQUIRED];
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto
@@ -17,6 +17,8 @@ syntax = "proto3";
 package wfa.measurement.api.v2alpha;
 
 import "google/api/field_behavior.proto";
+import "google/protobuf/struct.proto";
+import "wfa/measurement/type/future_disposition.proto";
 
 option java_package = "org.wfanet.measurement.api.v2alpha";
 option java_multiple_files = true;
@@ -50,4 +52,11 @@ message EventGroupMetadata {
     // Metadata for an [EventGroup][] containing ad impression events.
     AdMetadata ad_metadata = 2;
   }
+
+  // Metadata for the entity that the [EventGroup][] represents.
+  //
+  // The schema is specific to the ([DataProvider][],
+  // [EventGroup.EntityKey.entity_type][]) pair.
+  google.protobuf.Struct entity_metadata = 3
+      [(type.future_disposition) = REQUIRED];
 }
--- a/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/event_groups_service.proto
@@ -241,6 +241,11 @@ message ListEventGroupsRequest {
     // [EventGroup][]. [start_date][DateInterval.start_date] and
     // [end_date][DateInterval.end_date] are required.
     DateInterval activity_contains = 9;
+    // Set of entity types which [EventGroup.EntityKey.type][] of
+    // [EventGroup.entity_key][] must be in.
+    //
+    // If unspecified, the server will treat this as if it contains "campaign".
+    repeated string entity_type_in = 11;
   }
   // Filter criteria for this request.
   //


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 0 of 4 files reviewed, 1 unresolved discussion.

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq73j1u2k8aoYPloonb:-Oq73j1u2k8aoYPloonc:bwwv7wp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/beb6d72cf42ae9631ddccaa23c3253ba9b5e74c6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L242)):*
> ```proto
>   //
>   // Once specified, this may not change.
>   EntityKey entity_key = 16 [(type.future_disposition) = REQUIRED];
> ```

All existing EventGroups will have an entity_type of "campaign" for backwards-compatibility.

raimundoltdf (commented):
Looks good! I like the s/version/entity_type, and making entity_type a string to avoid dependencies on Halo releases. Do we need to replicate these changes in the internal API and the edpaggregator API?

> @raimundoltdf made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-:-Oq79cO96Tf9JkP9N4XV:b-svamr4)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on SanjayVas).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 231 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7ASbe1pXesCFEIdYs:-Oq7ASbe1pXesCFEIdYt:bjmk6uw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L231)):*
> ```proto
>     // ID of the entity in the [DataProvider][]'s system.
>     //
>     // This should be an ID that is available to [MeasurementConsumer][]'s
> ```

Is there anyway to enforce this? otherwise I'd make this statement more optional. "Ideally..."
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-Oq7A2PS8GjbUGEcTzrp:buyy274) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*
> ```proto
>     // This must be a URL-safe string, such that URL-encoding or -decoding
>     // results in the same string.
>     string entity_id = 2 [(type.future_disposition) = REQUIRED];
> ```

I'm assuming this is optional, is this unique (when set) across all the entity_ids or all (entity_ids, entity_type) pairs. Either way I'd document which.

SanjayVas (commented):
Yes, in the sense that we'd need to update the internal API to implement this and update the EDPA input format to include the new metadata and the entity key.

> @SanjayVas made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-:-Oq7SZ_PCcaCVzVud0_E:b-t568b5)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on raimundoltdf and SanjayVas).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 231 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7ASbe1pXesCFEIdYs:-Oq7Sx_f-EycbnMsjii8:b4wtdop) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L231)):*
<details><summary><i>Previously, raimundoltdf wrote…</i></summary><blockquote>

Is there anyway to enforce this? otherwise I'd make this statement more optional. "Ideally..."
</blockquote></details>

The "should" is already that. Otherwise this would say "must".

In either case there's no way to enforce it.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-Oq7TOeSDlDOq0TW4eox:b-mmprve) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*
<details><summary><i>Previously, raimundoltdf wrote…</i></summary><blockquote>

I'm assuming this is optional, is this unique (when set) across all the entity_ids or all (entity_ids, entity_type) pairs. Either way I'd document which.
</blockquote></details>

Yes, everything in the API that doesn't have `(google.api.field_behavior) = REQUIRED` is optional. This one instead has `(type.future_disposition) = REQUIRED` to indicate that we intend to make it required in the future, but it's not required yet.

I'll add the comment on uniqueness.

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-Oq8ZD_xBB-8_KycP9yG:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*

Yes, everything in the API that doesn't have `(google.api.field_behavior) = REQUIRED` is optional. This one instead has `(type.future_disposition) = REQUIRED` to indicate that we intend to make it required in the future, but it's not required yet.

I'll add the comment on uniqueness.
</blockquote></details>

Done.

SanjayVas (commented):
> @SanjayVas made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 0 of 4 files reviewed, 4 unresolved discussions (waiting on kungfucraig and raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq73j1u2k8aoYPloonb:-OqC67y4-wx9XLuJOQ7q:bd753zq) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/beb6d72cf42ae9631ddccaa23c3253ba9b5e74c6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L242)):*

All existing EventGroups will have an entity_type of "campaign" for backwards-compatibility.
</blockquote></details>

Potential implementation: column in Spanner with a DEFAULT value.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqC5LQVE5nBSTlKLJHv:-OqC5LQVE5nBSTlKLJHw:bs6jbjz) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*
> ```proto
>   // The schema is specific to the ([DataProvider][],
>   // [EventGroup.EntityKey.entity_type][]) pair.
>   google.protobuf.Struct entity_metadata = 3
> ```

We may want to use `map<string, string>` here instead to make it easier to do text search on the values, which for protobufs requires generated columns. See https://docs.cloud.google.com/spanner/docs/secondary-indexes#proto-indexing

Otherwise an alternative DB structure may be needed (i.e. not just persisting the Struct in a protobuf field).

@kungfucraig @raimundoltdf WDYT?

kungfucraig (commented):
> @kungfucraig reviewed 1 file and all commit messages, and made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 1 of 4 files reviewed, 4 unresolved discussions (waiting on raimundoltdf and SanjayVas).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq73j1u2k8aoYPloonb:-OqCAKlv0EMTqnU3dEk-:bob2i8g) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/beb6d72cf42ae9631ddccaa23c3253ba9b5e74c6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L242)):*

Potential implementation: column in Spanner with a DEFAULT value.
</blockquote></details>

I think the EDPs should be asked to update/delete their existing EGs as part of adding the enhanced metadata.

Use of this field without the EntityId is not useful.

We also don't know if the existing EGs actually represent campaigns.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-Oq7aMfQDNd_VS-gBg3i:b-rhyo3d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*

Done.
</blockquote></details>

The use of the entire EntityKey structure is already marked as future required.  So why not make this required to start with? Otherwise, it seems like it should just be marked optional since if someone can't use this field when they create their first EntityKey they won't ever use it.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqC5LQVE5nBSTlKLJHv:-OqCRuW6-O8VqwXSHIEv:b-y072x2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*

We may want to use `map<string, string>` here instead to make it easier to do text search on the values, which for protobufs requires generated columns. See https://docs.cloud.google.com/spanner/docs/secondary-indexes#proto-indexing

Otherwise an alternative DB structure may be needed (i.e. not just persisting the Struct in a protobuf field).

@kungfucraig @raimundoltdf WDYT?
</blockquote></details>

I see a need for repeated fields, so the struct is nice. However, we could define a convention for the map keys for this case.

I'm okay reverting to the map for indexing simplicity, but we need to be clear about how we want to interpret keys.

SanjayVas (commented):
> @SanjayVas made 3 comments and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq73j1u2k8aoYPloonb:-OqCa1uAAlZ-D43rEPQI:b-6tuacs) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/beb6d72cf42ae9631ddccaa23c3253ba9b5e74c6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L242)):*
We're basically saying that the API is assuming every existing EG is a campaign until the EDP states otherwise. This is an assumption that the API already makes since the only EG metadata that's currently supported is campaign, with no way to differentiate multiple EG's belonging to the same EDP campaign.

> I think the EDPs should be asked to update/delete their existing EGs as part of adding the enhanced metadata.

The delete should only be needed if the EG in question doesn't make sense - e.g. if it's a level in the hierarchy that is no longer being exported. I expect in practice all existing EGs can stay. Any existing UIs and API callers will get the existing EGs, and any new UI that supports the more granular EGs will get those. It effectively functions the same as the `metadata_version` field in the initial LTDF proposal.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-OqC_Lu7Dy3kO_B6uL2e:b-9vd90c) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*

The use of the entire EntityKey structure is already marked as future required.  So why not make this required to start with? Otherwise, it seems like it should just be marked optional since if someone can't use this field when they create their first EntityKey they won't ever use it.
</blockquote></details>

Adding a required field is a breaking change. EDPs won't have a chance to test their implementation of this against a real instance (e.g. QA) if it's immediately required. So the sequencing has to be

1. Add the field.
2. Get all EDPs to update their impls
3. Make the field required.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqC5LQVE5nBSTlKLJHv:-OqCbzcL3dCHMgSaKpvO:b-vvnfap) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*

I see a need for repeated fields, so the struct is nice. However, we could define a convention for the map keys for this case.

I'm okay reverting to the map for indexing simplicity, but we need to be clear about how we want to interpret keys.
</blockquote></details>

FWIW I prefer Struct. I just wanted to highlight that the DB schema may need to be a bit more complicated in that case in order to keep the door open for searchability. One solution may be to switch from substring search to JSON search, using a generated column to get a JSON version of the protobuf struct. See https://docs.cloud.google.com/spanner/docs/full-text-search/json-indexes

kungfucraig (commented):
> @kungfucraig made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq73j1u2k8aoYPloonb:-OqCgJMxA3auKpIXiUbI:b541e8d) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/beb6d72cf42ae9631ddccaa23c3253ba9b5e74c6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L242)):*

We're basically saying that the API is assuming every existing EG is a campaign until the EDP states otherwise. This is an assumption that the API already makes since the only EG metadata that's currently supported is campaign, with no way to differentiate multiple EG's belonging to the same EDP campaign.

> I think the EDPs should be asked to update/delete their existing EGs as part of adding the enhanced metadata.

The delete should only be needed if the EG in question doesn't make sense - e.g. if it's a level in the hierarchy that is no longer being exported. I expect in practice all existing EGs can stay. Any existing UIs and API callers will get the existing EGs, and any new UI that supports the more granular EGs will get those. It effectively functions the same as the `metadata_version` field in the initial LTDF proposal.
</blockquote></details>

I don't see a good reason to default this in the DB if we cannot also supply the entity ID. The API could provide a default if we want, but I don't see much value there either. Meaning, it's not worth the effort.

Having the EDPs make the update or delete these seems fine, since the real value is in having the entity ID.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-OqCfvWZ2GlI0vhjRuPK:b-39n0l7) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*

Adding a required field is a breaking change. EDPs won't have a chance to test their implementation of this against a real instance (e.g. QA) if it's immediately required. So the sequencing has to be

1. Add the field.
2. Get all EDPs to update their impls
3. Make the field required.
</blockquote></details>

But this is a subfield entity_key, which is the actual field added to the API. Moreover the EntityType above is marked required. At the very least entity_id should be consistent with entity_type.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqC5LQVE5nBSTlKLJHv:-OqChLmkEuGNZCrn0F0D:b-zd7tnp) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*

FWIW I prefer Struct. I just wanted to highlight that the DB schema may need to be a bit more complicated in that case in order to keep the door open for searchability. One solution may be to switch from substring search to JSON search, using a generated column to get a JSON version of the protobuf struct. See https://docs.cloud.google.com/spanner/docs/full-text-search/json-indexes
</blockquote></details>

That could work. I'd prefer to stick with the struct if the ability to search is not affected.

SanjayVas (commented):
> @SanjayVas made 3 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 1 of 4 files reviewed, 3 unresolved discussions (waiting on raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 242 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq73j1u2k8aoYPloonb:-OqDAkXQ0xnGw0doFEWJ:b3m5wz9) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/beb6d72cf42ae9631ddccaa23c3253ba9b5e74c6/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L242)):*

I don't see a good reason to default this in the DB if we cannot also supply the entity ID. The API could provide a default if we want, but I don't see much value there either. Meaning, it's not worth the effort.

Having the EDPs make the update or delete these seems fine, since the real value is in having the entity ID.
</blockquote></details>

We need to default this to avoid breaking things while both the existing coarse and new granular EGs exist.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 236 at r3](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-Oq7BCBjBLlGWBrDGgT6:-OqDB-M465_5vcVRdjOT:b7tidu2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/3612e1d5ee737c6cbf152109e252aa89ba385df9/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L236)):*

But this is a subfield entity_key, which is the actual field added to the API. Moreover the EntityType above is marked required. At the very least entity_id should be consistent with entity_type.
</blockquote></details>

See other comment re:defaulting. This means that initially we'll have entity_type but no entity_id on existing EGs.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqC5LQVE5nBSTlKLJHv:-OqDB2TTCVrkAvucF_IW:b-mr8lrw) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*

That could work. I'd prefer to stick with the struct if the ability to search is not affected.
</blockquote></details>

SGTM. An approach to make the existing substring work would be to just tokenize the JSON string.

Either way, I think I've convinced myself that just persisting a Struct protobuf message in the DB details won't make it harder to implement search later.

kungfucraig (approved):
> @kungfucraig reviewed 3 files and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: all files reviewed, 2 unresolved discussions (waiting on raimundoltdf).

stevenwarejones (approved):
> @stevenwarejones reviewed 4 files and all commit messages, and made 2 comments.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: all files reviewed, 4 unresolved discussions (waiting on raimundoltdf and SanjayVas).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 237 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqN5Ucu4Kk9hDsjVQhh:-OqN5Ucu4Kk9hDsjVQhi:b-xlctu2) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L237)):*
> ```proto
>     // results in the same string.
>     string entity_id = 2 [(type.future_disposition) = REQUIRED];
>   }
> ```

i'd also like an entity human readable name string field as well.
___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqN6F23DEFMuz260kYw:-OqN6F23DEFMuz260kYx:buferse) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*
> ```proto
>   // The schema is specific to the ([DataProvider][],
>   // [EventGroup.EntityKey.entity_type][]) pair.
>   google.protobuf.Struct entity_metadata = 3
> ```

is this where any edp would put any arbitrary metadata about the event group?

SanjayVas (commented):
> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: all files reviewed, 4 unresolved discussions (waiting on raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 237 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqN5Ucu4Kk9hDsjVQhh:-OqNKkTpBH9KsjwIoqZh:blr749b) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L237)):*

i'd also like an entity human readable name string field as well.
</blockquote></details>

The type is supposed to be human-readable, e.g. "campaign", "creative", "ad-group". I'll add a comment to make that clear.

SanjayVas (commented):
> @SanjayVas made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: all files reviewed, 3 unresolved discussions (waiting on raimundoltdf).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto` line 60 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqN6F23DEFMuz260kYw:-OqNKypVBZhook_2oEFS:b2nhswv) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group_metadata.proto#L60)):*

is this where any edp would put any arbitrary metadata about the event group?
</blockquote></details>

Yep

SanjayVas:
I'm going ahead and merging this despite not having an implementation ready yet to give EDPs a head start on integrating the changes. This is safe because there are no breaking changes.

SanjayVas (commented):
> @SanjayVas made 1 comment and resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on kungfucraig, raimundoltdf, and stevenwarejones).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/event_group.proto` line 237 at r4](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-OqN5Ucu4Kk9hDsjVQhh:-OqNMBKiDM5HY1VwrAEP:b-896fix) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/c1f9408163c449829665af0a67ab08d00c128b26/src/main/proto/wfa/measurement/api/v2alpha/event_group.proto#L237)):*

The type is supposed to be human-readable, e.g. "campaign", "creative", "ad-group". I'll add a comment to make that clear.
</blockquote></details>

Done.

SanjayVas (commented):
> @SanjayVas reviewed 1 file.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: all files reviewed, 2 unresolved discussions (waiting on kungfucraig and raimundoltdf).

SanjayVas (commented):
> @SanjayVas resolved 2 discussions and dismissed @raimundoltdf from 2 discussions.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on kungfucraig).

stevenwarejones (commented):
> @stevenwarejones reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

stevenwarejones (commented):
okay. where does the hierarchy information get stored in this?

> @stevenwarejones made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-:-OqNxXIN6CmV6IUjrOa5:b-h7157c)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

SanjayVas (commented):
Unfortunately it isn't known to the API. The `entity_metadata` would contain the denormalized metadata for all levels of the hierarchy. Each caller needs to know which keys map to which levels in the hierarchy for each EDP. That was the crux of the [design proposal](https://docs.google.com/document/d/1UJWPQUHMmKe_Rk2Xjetn-Tc7qqF9lKkhUr9NyKUQ9ck/edit?usp=sharing) and the discussion surrounding it: defer having the hierarchical information in the API until later.

> @SanjayVas made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/275#-:-OqRfIY_EkBOrHrF4Qak:budd4za)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved.

PR #293, opened 2026-08-05
Replace AwsKmsParams.credential_source with a token_audience oneof

Marco-Premier:
The credential source was redundant with the audience field, whose meaning it selected, and it relied on the unspecified enum value meaning GCP_WORKLOAD_IDENTITY. Modelling the two audiences as a oneof makes the credential source implicit in which member is set, so an inconsistent combination is unrepresentable and no zero value has to be reinterpreted.

Existing writers keep field 4, so their messages still read as the workload-identity variant. The Confidential Space audience takes field 6 rather than 5 because field 5 was released as the enum and has a different wire type.

Issue: https://github.com/world-federation-of-advertisers/cross-media-measurement/issues/4309

### DIFF
--- a/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto
+++ b/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto
@@ -112,27 +112,26 @@ message FulfillRequisitionRequest {
           string role_session = 2;
           // AWS region for STS and KMS endpoints.
           string region = 3;
-          // Audience presented to AWS STS when assuming `role_arn`: the GCP
-          // ID-token audience for `GCP_WORKLOAD_IDENTITY`, or the Confidential
-          // Space attestation-token audience for `CONFIDENTIAL_SPACE`.
-          string audience = 4;
 
-          // How the workload authenticates to AWS STS to assume `role_arn`.
-          enum CredentialSource {
-            // Unspecified; treated as `GCP_WORKLOAD_IDENTITY` for backward
-            // compatibility.
-            CREDENTIAL_SOURCE_UNSPECIFIED = 0;
-            // Two-hop: GCP Workload Identity Federation plus service-account
-            // impersonation mints the OIDC token presented to AWS STS. Requires
-            // `workload_identity_provider` and `impersonated_service_account`.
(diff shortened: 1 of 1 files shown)


wfa-reviewable:
This change is [Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/293)

SanjayVas (approved):
> @SanjayVas reviewed 1 file and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/293)* status: all files reviewed, 1 unresolved discussion (waiting on Marco-Premier).

___
*[`src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto` line 116 at r1](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/293#-OzIe7Ia7RoonEr4mdwk:-OzIe7Ia7RoonEr4mdwl:b-5y5xbb) ([raw file](https://github.com/world-federation-of-advertisers/cross-media-measurement-api/blob/b25b8933352504440434166e9a968f967bc30926/src/main/proto/wfa/measurement/api/v2alpha/requisition_fulfillment_service.proto#L116)):*
> ```proto
>           string region = 3;
> 
>           // Field 5 held a `credential_source` enum that the `token_audience`
> ```

nit: move the `reserved` option to the top of the message definition.

Typically it's a single `reserved` option with all the reserved tag numbers. e.g.

```proto
reserved 5, 10, 15;
```

We usually don't need comments explaining what the field used to be as it can be determined via commit history.

Marco-Premier (commented):
> @Marco-Premier resolved 1 discussion.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/293)* status: 0 of 1 files reviewed, all discussions resolved (waiting on SanjayVas).

Marco-Premier (commented):
> @Marco-Premier reviewed 1 file and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/world-federation-of-advertisers/cross-media-measurement-api/293)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on SanjayVas).

